PR #1, opened 2016-08-28
[Refactor] Create ManageIQ class and move all relevant methods inside it

dkorn:


dkorn:

library/manageiq.py
```diff
@@ -64,14 +61,86 @@
...
+                                      connection_configurations=endpoints)
+            provider_id = result['results'][0]['id']
+            self.msg = "Successfuly added %s provider" % provider_name
+            self.changed = True
+        except Exception as e:
+            self.msg = "Failed to add provider. Error: %s" % e
+            self.module.fail_json(msg=self.msg)
+        return provider_id
+
+    def add_provider(self, provider_name, hostname, port, token):
```
cben: `add_new_provider` and `add_provider` sound more or less equivalent.
Consider `add_or_update_provider` or maybe `set_provider` / `config_provider`?

library/manageiq.py
```diff
@@ -64,14 +61,86 @@
...
+    def update_required(self, provider_id, hostname, port):
+        try:
+            result = self.client.get(self.providers_url + '/%d/?attributes=authentications,endpoints' % provider_id)
+            endpoint = result['endpoints'][0]
+            return False if (endpoint['hostname'] == hostname and endpoint['port'] == int(port)) else True
+        except Exception as e:
+            self.msg = "Failed to get provider data. Error: %s" % e
+            self.module.fail_json(msg=self.msg)
+
+    def update_provider(self, endpoints, provider_name, provider_id):
```
cben: opposite arg order from `add_new_provider(provider_name, endpoints)`.
Personally I find (id, name, endpoints) order most natural, but either is good as long as it's consistent.

library/manageiq.py
```diff
@@ -64,14 +61,86 @@
...
+            provider_id = result['results'][0]['id']
+            self.msg = "Successfuly added %s provider" % provider_name
+            self.changed = True
+        except Exception as e:
+            self.msg = "Failed to add provider. Error: %s" % e
+            self.module.fail_json(msg=self.msg)
+        return provider_id
+
+    def add_provider(self, provider_name, hostname, port, token):
+        #adds an OpenShift containers provider to manageiq
```
cben: Please use docstring rather than comment.
This comment is misleading — it adds or updates.

Please document return value / possibility of exit.

library/manageiq.py
```diff
@@ -64,14 +61,86 @@
...
+            result = self.client.post(self.providers_url, type=ManageIQ.openshift_provider_type,
+                                      name=provider_name,
+                                      connection_configurations=endpoints)
+            provider_id = result['results'][0]['id']
+            self.msg = "Successfuly added %s provider" % provider_name
+            self.changed = True
+        except Exception as e:
+            self.msg = "Failed to add provider. Error: %s" % e
+            self.module.fail_json(msg=self.msg)
+        return provider_id
```
cben: Please add docstring documenting return value / possibility of exit.

library/manageiq.py
```diff
@@ -87,59 +156,17 @@ def main():
...
-            result = client.post(providers_url, type=PROVIDER_TYPE,
-                                 name=provider_name,
-                                 connection_configurations=endpoints)
-            provider_id = result['results'][0]['id']
-            msg = "Successfuly added %s provider" % provider_name
-            changed = True
-        except Exception as e:
-            msg = "Failed to add provider. Error: %s" % e
-            module.fail_json(msg=msg)
+    url           = module.params['url'] + '/api'
```
cben: consider calling this `api_url` to avoid different meanings between module's parameter and internal variable of same name.

library/manageiq.py
```diff
@@ -64,14 +61,86 @@
...
-    try:
-        result = client.get(providers_url + '/%d/?attributes=authentications,endpoints' % provider_id)
-        endpoint = result['endpoints'][0]
-        return False if (endpoint['hostname'] == hostname and endpoint['port'] == int(port)) else True
-    except Exception as e:
-        msg = "Failed to get provider data. Error: %s" % e
-        module.fail_json(msg=msg)
+class ManageIQ(object):
+    """
+        ManageIQ object to execute various operations in manageiq
```
cben: This description doesn't say much, but I don't have concrete improvement suggestion.

nitpick: align under """ (4 spaces not 8) please

cben:
Some nitpicks but overall improved :+1:

library/manageiq.py
```diff
@@ -64,14 +61,86 @@
...
+        self.changed       = False
+        self.msg           = ''
+        self.providers_url = self.url + '/providers'
+
+
+    def update_required(self, provider_id, hostname, port):
+        try:
+            result = self.client.get(self.providers_url + '/%d/?attributes=authentications,endpoints' % provider_id)
+            endpoint = result['endpoints'][0]
+            return False if (endpoint['hostname'] == hostname and endpoint['port'] == int(port)) else True
```
simon3z: @dkorn the above two lines shouldn't be in the try/except block.

library/manageiq.py
```diff
@@ -64,14 +61,86 @@
...
+        except Exception as e:
+            self.msg = "Failed to get provider data. Error: %s" % e
```
simon3z: @dkorn why do you store the message is the object attiribute? Is this a standard ansible technique? It seems weird.

library/manageiq.py
dkorn: I agree. I think I'll go with `add_or_update_provider` cause IMHO `set  / config` aren't explicit enough to the method actions

library/manageiq.py
dkorn: Agree, I'll go with (id, name, endpoints)

library/manageiq.py
dkorn: will do, thanks

library/manageiq.py
dkorn: thanks, will do

library/manageiq.py
dkorn: I agree, thanks

library/manageiq.py
dkorn: thanks

library/manageiq.py
dkorn: right, thanks!

library/manageiq.py
```diff
@@ -64,14 +61,86 @@
...
+                                   'port': port},
+                     'authentication': {'role': 'bearer',
+                                        'auth_key': token}}]
+
+        # check if provider with the same name already exists
+        providers = self.client.collections.providers
+        provider_id = None
+        for provider in providers:
+            if provider.name == provider_name:
+                provider_id = provider.id
```
cben: Consider extracting a find_provider_by_name method.

library/manageiq.py
```diff
@@ -64,14 +61,86 @@
...
+        if provider_id: #provider exists
+            if self.update_required(provider_id, hostname, port):
+                self.update_provider(endpoints, provider_name, provider_id)
```
cben: If you move building `endpoints` inside `update_provider`,
this function would be shorter and with nice symmetry between `update_required()` and `update_provider()` params.
Perhaps even move `update_required()` check inside `update_provider`?
[not necessarily in this PR]

cben:
BTW, https://docs.ansible.com/ansible/developing_modules.html#module-checklist mentions "For password / secret arguments `no_log=True` should be set."

library/manageiq.py
dkorn: @simon3z  it is standard to output a message as part of `module.exit_json` in case nothing failed (in case of failure the output is `module.fail_json`). the message is usually set in an inner method and stored as an attribute. 
I agree that in case of failure I could just do `msg =` and output `self.module.fail_json(msg=msg)` but I wanted to maintain consistency.

library/manageiq.py
dkorn: great advice, will do that.

library/manageiq.py
dkorn: > If you move building endpoints inside update_provider,
> this function would be shorter and with nice symmetry between update_required() and update_provider() params.

the thing is, `add_new_provider` also uses `endpoints`

> Perhaps even move update_required() check inside update_provider?
> [not necessarily in this PR]

perhaps in a separate PR, thanks!

dkorn:
> BTW, https://docs.ansible.com/ansible/developing_modules.html#module-checklist mentions "For password / secret arguments no_log=True should be set."

@cben Done, thanks!

library/manageiq.py
```diff
@@ -64,14 +61,110 @@
...
+        return provider_id
+
+    def find_provider_by_name(self, provider_name):
+        """ Searches the provider name in manageiq existing providers
+
+        Returns:
+            the provider id if it exists in manageiq, None otherwise
+        """
+        providers = self.client.collections.providers
+        for provider in providers:
```
enoodle: Suggestion: `next((p for p in providers if p.name == provider_name), None)`

library/manageiq.py
simon3z: @dkorn what consistency? Everywhere you do:

``` python
self.msg = "(message)"
self.module.fail_json(msg=self.msg)
```

Just change it to:

``` python
self.module.fail_json(msg="(message)")
```

enoodle:
LGTM

library/manageiq.py
```diff
@@ -64,14 +61,110 @@
...
+
+        Returns:
+            False if the hostname and port passed equals the provider's,
+            True otherwise
+        """
+        try:
+            result = self.client.get(self.providers_url + '/%d/?attributes=authentications,endpoints' % provider_id)
+        except Exception as e:
+            self.msg = "Failed to get provider data. Error: %s" % e
+            self.module.fail_json(msg=self.msg)
```
simon3z: @dkorn don't miss my previous comment about `self.msg` I continued the discussion on the previous code.

library/manageiq.py
dkorn: @enoodle I'll use it (just change to `p.id`), thanks!

library/manageiq.py
```diff
@@ -64,14 +61,110 @@
...
-    try:
-        result = client.get(providers_url + '/%d/?attributes=authentications,endpoints' % provider_id)
+class ManageIQ(object):
+    """ ManageIQ object to execute various operations in manageiq
+
+    url      - manageiq environment url
+    user     - the username in manageiq
+    password - the user password in manageiq
+    """
+    openshift_provider_type = 'ManageIQ::Providers::OpenshiftEnterprise::ContainerManager'
```
cben, line 71: [for later PR] Ideally this would be module param e.g.

```
  manageiq_provider:
    type: openshift
    name: My Cluster
    ...
```

Should be required, even if currently we support only one (or two) type — there's no sensible default.

library/manageiq.py
cben: Neat!  But even though I know how this works, I find the existing for: if: return code more readable.
EDIT: I not strongly against it, just not sure it's a win.  Daniel's call.

library/manageiq.py
cben: Agreed.
`fail_json()` does `sys.exit(1)`; setting an ivar gives false impression it matters ("is it used later?  does fail_json() return?"), so is slightly confusing to read.

cben:
LGTM (up to `self.msg` on fail paths)
:shipit:

dkorn:
@cben @simon3z @enoodle I fixed the `self.msg` thing, merging.
Thanks!

PR #2, opened 2016-08-30
Add Openshift provider type option to the module

dkorn:
The choices are: 
- 'openshift-origin'
- 'openshift-enterprise'

### DIFF
--- a/library/manageiq.py
+++ b/library/manageiq.py
@@ -13,37 +13,42 @@ author: Daniel Korn (dkorn)
 options:
   url:
     description:
-      - 'the manageiq environment url'
+      - the manageiq environment url
     required: true
     default: []
   username:
     description:
-      - 'manageiq username'
+      - manageiq username
     required: true
     default: []
   password:
     description:
-      - 'manageiq password'
+      - manageiq password
     required: true
     default: []
   name:
     description:
-      - 'the added provider name in manageiq'
+      - the added provider name in manageiq
     required: true
     default: []
+  type:
+    description:
+      - the openshift provider type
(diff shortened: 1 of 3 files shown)


dkorn:

library/manageiq.py
```diff
@@ -178,13 +187,14 @@ def main():
...
     provider_name = module.params['name']
+    provider_type = module.params['type']
     hostname      = module.params['hostname']
     port          = module.params['port']
     token         = module.params['token']
 
     manageiq = ManageIQ(module, url, username, password)
 
-    res_args = manageiq.add_or_update_provider(provider_name, hostname, port, token)
+    res_args = manageiq.add_or_update_provider(provider_name, provider_type, hostname, port, token)
```
cben, line 197: it may be time to start passing these k=v style

cben:
I think `type` should take `['openshift_origin', 'openshift_enterprise']`.
While you only care about openshift providers at this stage, in the future adding say Amazon, oVirt, etc. providers makes equal sense...

library/manageiq.py
dkorn: @cben why would that be useful? `add_or_update_provider` doesn't have default values

dkorn:
> I think type should take ['openshift_origin', 'openshift_enterprise'].
> While you only care about openshift providers at this stage, in the future adding say Amazon, oVirt, etc. providers makes equal sense...

I agree, done

cben:
LGTM

enoodle:
LGTM :+1:

PR #3, opened 2016-08-30
Add flake8 testing using tox.ini

jarovo:


dkorn:
@jaryn thanks for the contribution!
I plan to use travis to run flake8 on different python versions. 
Is tox better? adds a different functionality?

library/manageiq.py
```diff
@@ -97,9 +97,9 @@ def update_provider(self, provider_id, provider_name, endpoints):
         """ Updates the existing provider with new parameters
         """
         try:
-            result = self.client.post(self.providers_url + '/%d' % provider_id,
-                                      action='edit',
-                                      connection_configurations=endpoints)
+            self.client.post(self.providers_url + '/%d' % provider_id,
```
dkorn, line 100: thanks!

simon3z: @dkorn not for this PR but in general you would do something like: `'%s/%d' % (self.providers_url, provider_id)` or even better (python 3 style): `"{0}/{1}".format(self.providers_url, provider_id)`

dkorn: @simon3z I agree, will change all post api requests in a separate PR. thanks!

jarovo: Tox is a tool which serves multiple purposes. It can create virtualenvs, install deps and then run the tests. I don't know much about Travis, but Tox can be used to run the tests locally - on your computer. Travis may just run tox.

dkorn: @simon3z https://github.com/dkorn/manageiq-ansible-module/pull/5 handles the string formatting you suggested

tox.ini
```diff
@@ -0,0 +1,6 @@
+[tox]
+envlist = flake8
+
+[testenv:flake8]
+deps=flake8
+commands=flake8 --ignore=F403,E221,E501,F405 library
```
dkorn, line 9: @jaryn got it. I think the command should be: `commands=flake8 library/manageiq.py` ?
following your answer I'll merge this change and add a travis.yml.

jarovo: Well if the command looks like you are supposing there is a problem with that ansible modules code-style policy is more benevolent comparing to standard python (line length for example) so there will be many errors shown.

Another problem is that ansible modules require a star import which will cause many more flake8 complains (I hate ansible for this).

If done so, the resulting exit code of tox will be nonzero -- tests didn't pass.

If we use `library/manageiq.py`, we will test only the `manageiq.py` file even though there may be more of them in the future.

dkorn: > Well if the command looks like you are supposing there is a problem with that ansible modules code-style policy is more benevolent comparing to standard python (line length for example) so there will be many errors shown.
> 
> Another problem is that ansible modules require a star import which will cause many more flake8 complains (I hate ansible for this).
> 
> If done so, the resulting exit code of tox will be nonzero -- tests didn't pass.
> 

So, can we maybe ignore the specific errors ansible is more forgiving about, in addition to those you specified?

> If we use library/manageiq.py, we will test only the manageiq.py file even though there may be more of them in the future.

I agree :)

dkorn:
@jaryn 

> Well if the command looks like you are supposing there is a problem with that ansible modules code-style policy is more benevolent comparing to standard python (line length for example) so there will be many errors shown.
> 
> Another problem is that ansible modules require a star import which will cause many more flake8 complains (I hate ansible for this).
> 
> If done so, the resulting exit code of tox will be nonzero -- tests didn't pass.

So, can we maybe ignore the specific errors ansible is more forgiving about, in addition to those you specified?

> If we use library/manageiq.py, we will test only the manageiq.py file even though there may be more of them in the future.

I agree :)

jarovo:
@dkorn Yes we can. I would do it ad-hoc -- when we need to, we can always add a ignore statement. It is more hassle to remove it as it means reformatting the violating code.

Honestly, I would like the ansible not to be so benevolent with the code convention and I would LOVE it to not require using of the star-import.

dkorn:
@jaryn OK, I'll merge it and add the check as part of the travis.yml.
The star import appears in all ansible modules I checked so far, but investigate how to remove it and still use all the module functionalities.

jarovo:
You cannot remove it as ansible does search for that line and copy'n'pastes some code instead of it, then it sends that file trough scp to the destination where it executes the module.

Also beware that you cannot easily share code between the modules for same reason -- only one module is scp-ed to the destination.

PR #4, opened 2016-08-31
Add option to add provider with hawkular metrics

dkorn:
Adds the hawkular endpoints (endpoint and authentication). Also supports removing hawkular endpoints.

Need to specify:
- metrics: True (defualt is False)
- hawkular_hostname
- hawkular port

(the last two are required if metric is True)

### DIFF
--- a/library/manageiq.py
+++ b/library/manageiq.py
@@ -9,28 +9,28 @@ DOCUMENTATION = '''
 module: manageiq
 short_description: Execute various operations in ManageIQ
 requirements: [ ManageIQ/manageiq-api-client-python ]
-author: Daniel Korn (dkorn)
+author: Daniel Korn (@dkorn)
 options:
   url:
     description:
       - the manageiq environment url
     required: true
-    default: []
+    default: null
   username:
     description:
       - manageiq username
     required: true
-    default: []
+    default: null
   password:
     description:
       - manageiq password
     required: true
-    default: []
+    default: null
   name:
     description:
       - the added provider name in manageiq
     required: true
-    default: []
+    default: null
   type:
     description:
       - the openshift provider type
@@ -40,17 +40,33 @@ options:
     description:
       - the added provider hostname
     required: true
-    default: []
+    default: null
   port:
     description:
       - the port used by the added provider
     required: true
-    default: []
+    default: null
   token:
     description:
       - the added provider token
     required: true
-    default: []
+    default: null
+  metrics:
+    description:
+      - whether metrics should be enabled in the provider
+    required: false
+    default: False
+    choices: ['True', 'False']
+  hawkular_hostname:
+    description:
+      - the hostname used for hawkular metrics
+    required: false
+    default: null
+  hawkular_port:
+    description:
+      - the port used for hawkular metrics
+    required: false
+    default: null
 '''
 
 EXAMPLES = '''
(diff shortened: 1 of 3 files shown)


dkorn:

library/manageiq.py
```diff
@@ -91,7 +110,20 @@ def update_required(self, provider_id, hostname, port):
...
+                h_endpoint = result['endpoints'][1]
+                if (h_endpoint['hostname'] == endpoints[1]['endpoint']['hostname'] and
+                    h_endpoint['port'] == int(endpoints[1]['endpoint']['port'])):
+                    return False
+                else:
+                    return True
+            else:
+                return False
+        else:
+            return True
```
simon3z: @dkorn it seems that you should change the logic in order to have less nested if's.

library/manageiq.py
```diff
@@ -129,24 +161,41 @@ def find_provider_by_name(self, provider_name):
...
+
+        Returns:
+            the provider's endpoints, including the hawkular ones if metrics is required
+        """
+        endpoints = [{'endpoint': {'role': 'default', 'hostname': hostname,
+                                   'port': port},
+                      'authentication': {'role': 'bearer', 'auth_key': token}}]
+
+        # add hawkular endpoints if metrics is True
+        if self.module.params['metrics']:
```
simon3z, line 173: @dkorn can `metrics` be missing? If you use `self.module.params.get('metrics')`, then you would receive `None` in case `metrics` is not defined.

library/manageiq.py
dkorn: @simon3z I refactored it with the help of @cben, now comparing dictionaries.

library/manageiq.py
dkorn: @simon3z If `metrics` is missing the default value is `False`. 
I discovered that if `module.params['key']` doesn't exist the default value is None (therefore no need for the `.get()`

library/manageiq.py
```diff
@@ -90,8 +113,14 @@ def update_required(self, provider_id, hostname, port):
...
             self.module.fail_json(msg="Failed to get provider data. Error: %s" % e)
-        endpoint = result['endpoints'][0]
-        return False if (endpoint['hostname'] == hostname and endpoint['port'] == int(port)) else True
+
+        def host_port(endpoint):
+            return (endpoint['hostname'], int(endpoint['port']))
+
+        desired_by_role = {e['endpoint']['role']: host_port(e['endpoint']) for e in endpoints}
+        result_by_role = {e['role']: host_port(e) for e in result['endpoints']}
+        mypdb.set_trace()
```
enoodle: nice dict comprehensions, but you have a `pdb.set_trace()` here ;-)

dkorn: nice catch :) thanks!

dkorn:
@simon3z @enoodle any other comments?

library/manageiq.py
```diff
@@ -97,8 +116,13 @@ def update_required(self, provider_id, hostname, port):
             result = self.client.get(self.providers_url + '/%d/?attributes=authentications,endpoints' % provider_id)
         except Exception as e:
             self.module.fail_json(msg="Failed to get provider data. Error: %s" % e)
-        endpoint = result['endpoints'][0]
-        return False if (endpoint['hostname'] == hostname and endpoint['port'] == int(port)) else True
+
+        def host_port(endpoint):
```
enoodle: suggestion: use lambda function instead of a real inline function:
`host_port = lambda endpoint: (endpoint['hostname'], int(endpoint['port']))`

dkorn: @enoodle cool, I'll use it :) thanks!

dkorn:
@simon3z can I merge this one?

simon3z:
> @simon3z can I merge this one?

@dkorn couldn't review in detail, but it looks good :+1:

PR #5, opened 2016-08-31
Refactor string formatting in rest api requests

dkorn:


dkorn:

enoodle:
I like named substitutions better but I guess its a matter of preference. 
If you want look at the 2nd and 3rd examples here: https://docs.python.org/2/library/string.html#format-examples
Anywy the change LGTM :+1:

dkorn:
@simon3z your opinion?

simon3z:
@dkorn since these are urls you can look if by any chance you can use [urlparse](https://docs.python.org/2/library/urlparse.html), which is not used only to parse urls but also to construct them.

Other options would be:

``` python
self.client.post('/'.join((self.providers_url, provider_id)), ...)
```

cben:
I'd even drop self.providers_url and spell out
`'{api_url}/providers/{id}'.format(api_url=self.api_url, id=provider_id)`
with the goal of being clear at a glance what the url looks like.

Q: can't the client help us construct urls?  Can we use higher-level interface than .get / .post?
(Or are you deliberately sticking to them for stability when client evolves?  Do you even get any benefit over raw `urllib` / `requests`?)
Improving the client might be more productive than bikeshedding this code...

LGTM either way.

cben:
OK, tiny bikeshedding: if you keep .format with positional args, you can use `{} {}` instead of `{0} {1}`.

The `{0}` numbers don't help reader match to arguments as those are unnumbered; they're mainly useful for out-of-order scenarios like i18n (which are generally better served by named args, but translators might have no control of that).

EDIT: both my comments are perfect examples why bikeshedding is silly.  Sorry for wasting your time.  :shipit:

dkorn:
I decided to go with @cben and @enoodle's suggestion of named args. @simon3z IMHO urlparse is a bit overkill here, at least at this point.

@cben - you raised an interesting point that worth a discussion, I just think it isn't in the scope of this PR.

enoodle:
LGTM :+1:

PR #6, opened 2016-09-01
Fix Openshift provider type argument choices

dkorn:
Following a change in the Openshift type strings,a change in the strings of the module arguments choices is also needed


dkorn:
@cben really small fix, please review so I can merge

cben:
LGTM

PR #7, opened 2016-09-01
Add support for getting mangeiq url, username and password from the env vars by default

dkorn:
To use this option just add the vars to your env:
$ export MIQ_URL=http://localhost:3000
$ export MIQ_USERNAME=admin
$ export MIQU_PASSWORD=******

and remove them from the extra-vars string, when running the playbook

### DIFF
--- a/library/manageiq.py
+++ b/library/manageiq.py
@@ -1,5 +1,6 @@
 #!/usr/bin/python
 
+import os
 from ansible.module_utils.basic import *
 from miqclient.api import API as MiqApi
 
@@ -14,18 +15,15 @@ options:
   url:
     description:
       - the manageiq environment url
-    required: true
-    default: null
+    default: MIQ_URL env var if set. otherwise, it is required to pass it
   username:
     description:
       - manageiq username
-    required: true
-    default: null
+    default: MIQ_USERNAME env var if set. otherwise, it is required to pass it
   password:
     description:
       - manageiq password
-    required: true
-    default: null
+    default: MIQ_PASSWORD env var if set. otherwise, it is required to pass it
   name:
     description:
       - the added provider name in manageiq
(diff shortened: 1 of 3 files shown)


dkorn:
@kwoodson @enoodle @cben @simon3z please review

kwoodson:
@dkorn,

These changes look good.

Please add type='<int|str|list|dict>' to your module params.

cben:
Is it common for ansible modules to directly depend on env vars?
I can see how it's convenient, but I would expect a module to only depend on what the playbook is passing.

Can you do this instead in add_provider.yml?
https://docs.ansible.com/ansible/faq.html#how-do-i-access-shell-environment-variables

kwoodson:
@cben,

It is pretty common to support environment variables in ansible modules.  This allows for full flexibility when calling the modules.

Here is an ansible module for ec2.
http://docs.ansible.com/ansible/ec2_module.html

Look at the values for aws_access_key and aws_secret_key.  Default to env vars if set and if not, they must be passed.  This is the best of both worlds.

library/manageiq.py
```diff
@@ -214,9 +212,9 @@ def main():
             name=dict(required=True),
             type=dict(required=True,
                       choices=['openshift-origin', 'openshift-enterprise']),
-            url=dict(required=True),
-            username=dict(required=True),
-            password=dict(required=True, no_log=True),
+            url=dict(default=os.environ.get('MIQ_URL', None)),
+            username=dict(default=os.environ.get('MIQ_USERNAME', None)),
+            password=dict(default=os.environ.get('MIQ_PASSWORD', None)),
```
cben, line 217: Don't you also need some logic to enforce "otherwise, it is required to pass it"?

```
if password is None:
    fail_json(...)
```

etc.
Or maybe something like:

```
def env_var_or_required(var):
    if var in os.environ:
        return dict(default=os.environ[var])
    else:
        return dict(required=True)

...
    url=env_var_or_required('MIQ_URL'),
    username=env_var_or_required('MIQ_USERNAME'),
    password=env_var_or_required('MIQ_PASSWORD'),
```

kwoodson: Adding validation is a possibility.  I'd be ok with that approach.

I imagine this call in the constructor to ManageIQ would throw an exception to the fact that credentials were invalid or not found.
`
            self.client        = MiqApi(self.api_url, (self.user, self.password))
`

dkorn: > I imagine this call in the constructor to ManageIQ would throw an exception to the fact that credentials were invalid or not found.
>  self.client = MiqApi(self.api_url, (self.user, self.password))

@kwoodson Unfortunately, this validation isn't part of the api client constructor.
I added a separate validation for these 3 credentials.

cben:
LGTM

dkorn:
@kwoodson are you OK with the validation?

PR #9, opened 2016-09-04
Updating a Provider: report the updated values

enoodle:
Change `update_requiered` to return a hash of the values that need
updating and add that hash to the debug message.

```
ok: [localhost] => {
    "result": {
        "changed": true, 
        "msg": "Successfuly updated oshift01 provider. Changes: {'Removed': {u'hawkular': {'hostname': u'alon_gold.com', 'port': 1234}}, 'Updated': {'default': {'hostname': 'erez_oshift01.com'}}, 'Added': {}}", 
        "provider_id": 35
    }
}
```

### DIFF
--- a/library/manageiq.py
+++ b/library/manageiq.py
@@ -103,12 +103,18 @@ class ManageIQ(object):
         self.changed       = False
         self.providers_url = self.api_url + '/providers'
 
-    def update_required(self, provider_id, endpoints):
+    def required_updates(self, provider_id, endpoints):
         """ Checks whether an update is required for the provider
 
         Returns:
-            False if the hostname and port passed equals the provider's,
-            True otherwise
+            Empty Hash (None) - If the hostname and port passed equals the
+                                provider's current values
+            Hash of Changes   - Changes that need to be made if any endpoint is
+                                different than the current values of the
+                                provider. The hash will have three entries:
+                                    Updated, Removed, Added
+                                that will contain all the changed endpoints
+                                and their values.
         """
         try:
             result = self.client.get('{providers_url}/{id}/?attributes=authentications,endpoints'.format(providers_url=self.providers_url, id=provider_id))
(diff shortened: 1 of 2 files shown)


dkorn:
Thanks for the contribution @enoodle!
Perhaps it's worth changing the name of `update_requiered`, as it no longer returns a boolean value. I would suggest `required_updates`.
Also, is it possible to output **only** the attribute that was changed, rather than the whole endpoint?

enoodle:
@cben PTAL

cben:
I want to try making it simpler with sets intersection/difference.

Meta-concern: we're selling a simulation of what we expect MIQ to change, not what it _actually_ changed.  If there is some bug, it'll be confusing.  Perhaps it's worth doing a GET after the changed to confirm actual diff?

cben:
Also, I believe tuple keys are not valid json.

enoodle:
@cben Do you want to separate the keys with sets and then create the hashes? 
Also, I updated the output to be more json compliant and nicer too.

library/manageiq.py
```diff
@@ -118,11 +118,21 @@ def update_required(self, provider_id, endpoints):
...
+            return {}
+        updated = {role: {k: v for k, v in ep.items()
+                          if k not in result_by_role[role] or v != result_by_role[role][k]}
+                   for role, ep in desired_by_role.items()
+                   if role in result_by_role and ep != result_by_role[role]}
+        added = {role: ep for role, ep in desired_by_role.items()
+                 if role not in result_by_role}
+        removed = {role: ep for role, ep in result_by_role.items()
+                   if role not in desired_by_role}
+        return {"Updated": updated, "Added": added, "Removed": removed}
```
cben, line 139: I'm thinking something like:

```
result_keys = set(result_by_role)
desired_keys = set(desired_by_role)
added = sorted(desired_keys - result_keys)
removed = sorted(result_keys - desired_keys)
updated = sorted(role for role in (desired_keys & result_keys)
                 if desired_by_role[role] != result_by_role[role])
```

This omits the host port details, but I'm not sure how useful those are.

enoodle: IIUC I will then have to create the current hashes anyway with the keys coming from these sets instead of calculating it now inside the dict comprehension.

dkorn: @kwoodson is it important for you to know the values of the updated/added/removed provider attributes or only the keys?

cben:
LGTM either way.

I couldn't resist playing with recursive dict diffing some more:
https://github.com/cben/manageiq-ansible-module/blob/required_updates_experiments/required_updates_experiments.ipynb
but can't say it's better.

library/manageiq.py
kwoodson: @dkorn seeing changes is always good.

We met with the ansible team yesterday and they mentioned that modules can support a diff mode that only returns differences so I think this is good practice.

cben (approved):

library/manageiq.py
```diff
@@ -105,7 +105,7 @@ def __init__(self, module, url, user, password):
         self.changed       = False
         self.providers_url = self.api_url + '/providers'
 
-    def update_required(self, provider_id, endpoints):
+    def required_updates(self, provider_id, endpoints):
         """ Checks whether an update is required for the provider
```
dkorn, line 107: @enoodle can you please update the docstring to match the changes you added?

dkorn:
@enoodle you also need to change `msg` in `test_will_update_provider_if_present`
https://github.com/dkorn/manageiq-ansible-module/blob/master/tests/test_container_provider.py#L95

PR #10, opened 2016-09-05
Add first unittests

jarovo:


tox.ini
```diff
@@ -1,9 +1,23 @@
...
-commands=flake8 --ignore=F403,E221,E501,F405 library
+commands =
+	flake8 --ignore=F403,E221,E501,F405 library
+	flake8 tests setup.py
+
+[testenv]
+deps = mock
+       pytest
+       ansible
+       git+https://github.com/ManageIQ/manageiq-api-client-python.git
```
cben, line 21: Nice.
I wonder if we should define deps in setup.py to make it easier to install/run/test locally.

tests/test_container_provider.py
```diff
@@ -0,0 +1,121 @@
...
+    yield endpoints
+
+
+class AnsibleModuleFailed(Exception):
+    pass
+
+
+@pytest.fixture()
+def miq(miq_api_class, miq_ansible_module, the_provider):
+    miq_ansible_module.params = {'metrics': True}
```
cben, line 43: This would ideally be inside test cases that need it.
But currently IIUC they all need it, as `generate_endpoints()` peeks inside `self.module.params`, which should be refactored.

tests/test_container_provider.py
```diff
@@ -0,0 +1,121 @@
...
+         'endpoint': {'hostname': 'The HHostname',
+                      'port': '54321',
+                      'role': 'hawkular'}}
+    ]
+
+
+def test_generate_endpoints_without_metrics(miq):
+    miq.module.params = {'metrics': False}
+    endpoints = miq.generate_endpoints("The hostname", "12345", "The token",
+                                       object(), object())
```
cben, line 78: another case where tests reveal the code being suboptimal :-)
I think in actual usage these 2 params are None when unspecified in playbook.

cben:
Code looks great :+1:
I propose merging these ASAP, so future PRs can be tested.

tests/test_container_provider.py
jarovo: Yeah. With the `object()` I can actually test whether the `generate_endpoints` doesn't by accident put the non-None into the generated endpoints when no metrics was required.

tox.ini
jarovo: For testing locally just run `tox`

As far as the `manageiq-api-client-python` is not on pip we cannot put the deps to the `setup.py`:

```
error in manageiq-ansible-module setup command: 'install_requires' must be a string or list of strings containing valid project/version requirement specifiers; Invalid requirement, parse error at "'+https:/'"
```

.travis.yml
```diff
@@ -2,9 +2,7 @@ language: python
...
   - "2.7"
   - "3.4"
-env:
-  - TOX_ENV=flake8
 install:
-  - pip install tox
+  - pip install tox-travis
 script:
-  - tox -e $TOX_ENV
+  - tox
```
jarovo, line 8: I hope these changes are gonna work with Travis. I am a Travis newbie.

dkorn: can you please change `tox` to `tox -- -v`?

tests/test_container_provider.py
cben: Addressed in #15

tests/test_container_provider.py
```diff
@@ -0,0 +1,121 @@
...
+    miq_api_class.return_value.post.return_value = dict(results=[dict(id=0)])
+    miq_api_class.return_value.get.return_value = dict(endpoints=[
+        {'port': '12345', 'role': 'default', 'hostname': 'The hostname'}])
+    miq_api_class.return_value.collections.providers = [the_provider]
+
+    yield miq
+
+
+def test_generate_endpoints_with_metrics(miq):
+    endpoints = miq.generate_endpoints("The hostname", "12345", "The token",
```
dkorn, line 61: suggestion: can you maybe give these strings meaningful names, like:
`"openshift01.example.com"`, `"8443"` (or a random number) etc.

also, maybe store these strings in a constant since you're using them several times

tests/test_container_provider.py
dkorn: @cben great, thanks!
@jaryn maybe add a TODO here

tests/test_container_provider.py
cben: I volunteer addressing this in followup PR if you merge this as is.
I'm biased because I'm too lazy to rebase #15 ;-) 
but generally it's more fun to refactor tests once we have CI in place running them for every change.  :green_apple:

cben:
For reference, here is how this looks on Travis:
https://travis-ci.org/cben/manageiq-ansible-module/builds/158478664

tests/test_container_provider.py
```diff
@@ -0,0 +1,121 @@
...
+    miq_ansible_module.params = {'metrics': True}
+
+    def fail(msg):
+        raise AnsibleModuleFailed(msg)
+
+    miq_ansible_module.fail_json = fail
+    miq = manageiq.ManageIQ(miq_ansible_module, "http://themanageiq.tld",
+                            "The username", "The password")
+
+    miq_api_class.return_value.post.return_value = dict(results=[dict(id=0)])
```
dkorn, line 52: a. do we want to return more than the id of the added provider in the tests? not required at this step anyway. this is a reponse example:

```
[{'api_version': u'v1',
               'created_on': u'2016-09-08T15:11:29Z',
               'guid': u'84a4084a-75d6-11e6-92bc-0242ee817803',
               'id': 134,
               'name': u'Openshift01',
               'tenant_id': 1,
               'type': u'ManageIQ::Providers::Openshift::ContainerManager',
               'updated_on': u'2016-09-08T15:11:29Z',
               'zone_id': 1}]}
```

b. the update provider operation is also done with post request, so when we'll add tests for updating a provider maybe we'll need a way to differentiate between the two operations.also, maybe we could look at the `'updated_on'` key

tests/test_container_provider.py
dkorn: OK @cben, as long as you also change this https://github.com/dkorn/manageiq-ansible-module/pull/10/files#r78011990 so that we'll have more details about the tests run.

PR #12, opened 2016-09-06
Delete provider functionality and 'state' option

dkorn:
The 'state' option has two choices:
- 'present' - adds the provider if it does not exist or update the provider if the associated data is different
- 'absent' - deletes the provider if it exists

Delete action is done asynchronously as it could take a while to complete. Therefore, The delete task id is returned so it can be queried if needed.

### DIFF
--- a/library/manageiq.py
+++ b/library/manageiq.py
@@ -34,6 +34,15 @@ options:
       - the openshift provider type
     required: true
     choices: ['openshift-origin', 'openshift-enterprise']
+  state:
+    description:
+      - the state of the provider
+      - On present, it will add the provider if it does not exist or update the
+      provider if the associated data is different
+      - On absent, it will delete the provider if it exists
+    required: false
+    choices: ['present', 'absent']
+    default: 'present'
   hostname:
     description:
       - the added provider hostname
@@ -72,6 +81,7 @@ EXAMPLES = '''
   manageiq:
     name: 'Molecule'
     type: 'openshift-enterprise'
+    state: 'present'
     url: 'http://localhost:3000'
     username: 'admin'
     password: '******'
@@ -165,13 +175,36 @@ class ManageIQ(object):
                              'port': port},
                 'authentication': {'role': authtype, 'auth_key': token}}
 
+    def delete_provider(self, provider_name):
+        """ Deletes the provider
+
+        Returns:
+            the delete task id if a task was generated, whether or not
+            a change took place and a short message describing the operation
+            executed.
+        """
+        provider_id = self.find_provider_by_name(provider_name)
+        if provider_id:
+            try:
+                url = '{providers_url}/{id}'.format(providers_url=self.providers_url, id=provider_id)
+                result = self.client.post(url, action='delete')
+                if result['success']:
+                    self.changed = True
+                    return dict(task_id=result['task_id'], changed=self.changed, msg=result['message'])
+                else:
+                    return dict(task_id=None, changed=self.changed, api_error=result, msg="Failed to delete {provider_name} provider".format(provider_name=provider_name))
+            except Exception as e:
+                self.module.fail_json(msg="Failed to delete {provider_name} provider. Error: {error}".format(provider_name=provider_name, error=e))
+        else:
+            return dict(task_id=None, changed=self.changed, msg="Provider {provider_name} doesn't exist".format(provider_name=provider_name))
+
     def add_or_update_provider(self, provider_name, provider_type, endpoints):
         """ Adds an OpenShift containers provider to manageiq or update it's
         attributes in case a provider with the same name already exists
 
         Returns:
-            the added or updated provider id, whether or not a change took place
-            and a short message describing the operation executed
+            the added or updated provider id, whether or not a change took
+            place and a short message describing the operation executed
         """
         message = ""
         # check if provider with the same name already exists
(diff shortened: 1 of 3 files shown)


dkorn:

simon3z:
> @enoodle @cben @simon3z please review

@dkorn shouldn't Kenny review these as well?

enoodle:
@dkorn I see that there is no verification that the provider was deleted successfully by manageiq, is it not required anymore?

dkorn:
> @dkorn shouldn't Kenny review these as well?

right. @kwoodson can you review?

dkorn:
> @dkorn I see that there is no verification that the provider was deleted successfully by manageiq, is it not required anymore?

the deletion can take time usually, if there's a problem connecting to the provider. I'm still looking into this. Currently, I'm outputing that the deletion started (the manageiq response) and the id of the task handling it (so it can be queried):

```
"changed": true, 
"msg": "Provider id:126 name:'Openshift01' deleting", 
"task_id": 32
```

I can add a query for that task and wait for it some time, I'm just not sure what would be the best outcome if the task wasn't completed during that time. 
What do you think @kwoodson @simon3z @cben

enoodle:
@dkorn :+1: I was just curious. It doesn't seem important to me as I don't see the use case of it. We assume ManageIQ works well and will delete the task eventually and probably the user of the ansible module don't care anymore about this provider or he would have edited it instead of marking it for deletion.

library/manageiq.py
```diff
@@ -180,13 +191,43 @@ def generate_endpoints(self, hostname, port, token, h_hostname, h_port):
...
+        if provider_id:
+            try:
+                result = self.client.post('{providers_url}/{id}'.format(providers_url=self.providers_url, id=provider_id),
+                                          action='delete')
+                if result['success']:
+                    message = result['message']
+                    task_id = result['task_id']
+                    self.changed = True
+                else:
+                    message = "Failed to delete {provider_name} provider".format(provider_name=provider_name)
```
cben: do we get any more info from API when result['success'] is false but there wasn't exception?

library/manageiq.py
```diff
@@ -180,13 +191,43 @@ def generate_endpoints(self, hostname, port, token, h_hostname, h_port):
...
+        task_id = None
+        provider_id = self.find_provider_by_name(provider_name)
+        if provider_id:
+            try:
+                result = self.client.post('{providers_url}/{id}'.format(providers_url=self.providers_url, id=provider_id),
+                                          action='delete')
+                if result['success']:
+                    message = result['message']
+                    task_id = result['task_id']
+                    self.changed = True
```
cben, line 192: Do you need `self.changed` instance var to outlive this method?
I also don't like the long flow `message = ""; task_id = None` — setting them here — `res_args =...; return res_args`.

I'd immediately `return dict(task_id=result['task_id'], changed=True, msg=result['message'])` here,
and separate `return dict(changed=False, msg="... doesn't exist"...)` at the end.

library/manageiq.py
```diff
@@ -180,13 +191,43 @@ def generate_endpoints(self, hostname, port, token, h_hostname, h_port):
...
+                else:
+                    message = "Failed to delete {provider_name} provider".format(provider_name=provider_name)
+            except Exception as e:
+                self.module.fail_json(msg="Failed to delete {provider_name} provider. Error: {error}".format(provider_name=provider_name, error=e))
+        else:
+            message = "Provider {provider_name} doesn't exist".format(provider_name=provider_name)
```
cben: in this case I think `self.changed` will crash in following line.
I'd do independent return here, see [comment above](https://github.com/dkorn/manageiq-ansible-module/pull/12/files#r77831147)

library/manageiq.py
dkorn: I I haven't seen a case where result['success'] is false. 
If there's an error then the module fails with the full error, including message.

library/manageiq.py
dkorn: @cben in this specific case I don't have to set `self.changed` but as an instance var, used in `add_update_provider` and the methods it uses, I thought that consistency is nicer -> I think always keeping `self.changed` updated to the change status is important. 

regarding `message`, `task_id`, and the returns I agree. let me know if the change I did is what you had in mind.

library/manageiq.py
dkorn: why would if failover `self.changed`? see my new change.

library/manageiq.py
```diff
@@ -180,13 +191,36 @@ def generate_endpoints(self, hostname, port, token, h_hostname, h_port):
...
+
+        Returns:
+            the delete task id if a task was generated, whether or not
+            a change took place and a short message describing the operation
+            executed
+        """
+        provider_id = self.find_provider_by_name(provider_name)
+        if provider_id:
+            try:
+                result = self.client.post('{providers_url}/{id}'.format(providers_url=self.providers_url, id=provider_id),
```
cben: very long line.  split out `url = ...`, or wrap somehow

library/manageiq.py
```diff
@@ -180,13 +191,36 @@ def generate_endpoints(self, hostname, port, token, h_hostname, h_port):
...
+                                          action='delete')
+                if result['success']:
+                    self.changed = True
+                    return dict(task_id=result['task_id'], changed=self.changed, msg=result['message'])
+                else:
+                    return dict(task_id=None, changed=self.changed, msg="Failed to delete {provider_name} provider".format(provider_name=provider_name))
```
cben: consider including `api_error=result` in the dict, in hope it might contain some explanation why it failed.

cben:
You're right, I forgot self.changed being initialized and used elsewhere.
LGTM.

library/manageiq.py
dkorn: added, thanks

library/manageiq.py
dkorn: changed, thanks!

PR #13, opened 2016-09-07
Add miq prefix to manageiq related options: url, username and password

dkorn:


dkorn:

simon3z:
Can't you use a dictionary?

dkorn:
> Can't you use a dictionary?

@simon3z what do you mean by using a dictionary?

simon3z:
@dkorn instead of:

```
manageiq:
    miq_url: '{{ miq_url | default(omit) }}'
    miq_username: '{{ miq_username | default(omit) }}'
    miq_password: '{{ miq_password | default(omit) }}'
    ...
```

you would do:

```
manageiq:
    server:
        url: '{{ miq_url | default(omit) }}'
        username: '{{ miq_username | default(omit) }}'
        password: '{{ miq_password | default(omit) }}'
    ...
```

dkorn:
@simon3z got it. I actually looked at the zabbix modules and IINM this isn't a common practice for ops team.
example: https://github.com/ansible/ansible-modules-extras/blob/f953d5dc0c1418dcdc5db72e8fb630b6aa83997f/monitoring/zabbix_maintenance.py#L40

I can make the change but maybe @kwoodson can give us his opinion.

cben:
LGTM on the code, no opinion on the interface

cben:
I saw core modules tend to have many separate modules eg. ec2_vol,
ec2_group etc:
http://docs.ansible.com/ansible/list_of_cloud_modules.html#amazon

So maybe this will be manageiq_provider module?

kwoodson:
@cben, @dkorn 

+1 for the manageiq_provider for name.

dkorn:
@cben @kwoodson I agree regarding the separation of the module into:
`manageiq_provider` and `manageiq_policy` (as a first step at least).

@simon3z If we'll go with the separate modules approach I really don't see an important reason for the `server` namespace, as there won't be many more options. Also, this isn't very common for ansible modules, at least those I looked into.

If there are no other comments I'm merging this change.

simon3z:
> @simon3z If we'll go with the separate modules approach I really don't see an important reason for the server namespace, as there won't be many more options.

@dkorn the need for that will become clearer when you'll implement more modules. Feel free to merge (but I have the feeling this will change again going forward).

PR #15, opened 2016-09-08
Don't peek into module.params in ManageIQ class

cben:
Based on Jaroslav's #10 tests FTW.

Silly `generate_endpoints` refactoring, because it was bugging me.
The function now returns one endpoint, and the "if metrics" logic went into main().
This "if metrics" is now untested.  I could add main() unittests but I think if we want to test main it's more productive to test with an actual playbook.

Dropped check for missing hawkular host,port when metrics=True, verified ansible's `required_if` enforces this for us.

PR #16, opened 2016-09-08
Added testing

jarovo:
I've addressed the review comments.

### DIFF
--- a/tests/test_container_provider.py
+++ b/tests/test_container_provider.py
@@ -1,3 +1,4 @@
+# -*- coding: utf-8 -*-
 import pytest
 from mock import Mock
 
@@ -5,6 +6,16 @@ from ansible.module_utils.basic import AnsibleModule
 
 from miqclient.api import API
 import manageiq
+
+
+PROVIDER_NAME = "Provider name 1 with some unicode characters «ταБЬℓσ»"
+PROVIDER_HOSTNAME = "some-provider-hostname.tld"
+PROVIDER_TOKEN = "THE_PROVIDER_TOKEN"
+PROVIDER_PORT = 8443
+PROVIDER_ID = 134
+HAWKULAR_HOSTNAME = "some-hawkular-hostname.tld"
+HAWKULAR_PORT = 443
+MANAGEIQ_HOSTNAME = "http://themanageiq.tld"
 
 
 @pytest.fixture(autouse=True)
@@ -17,7 +28,7 @@ def miq_api_class(monkeypatch):
 @pytest.fixture
 def the_provider():
     the_provider = Mock()
-    the_provider.name = "The provider"
+    the_provider.name = PROVIDER_NAME
     yield the_provider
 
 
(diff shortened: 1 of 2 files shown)


tests/test_container_provider.py
```diff
@@ -7,6 +8,16 @@
 import manageiq
 
 
+provider_name = "Provider name 1 with some unicode characters «ταБЬℓσ»"
+provider_hostname = "some-provider-hostname.tld"
```
dkorn: shouldn't these constants be all capital letters?

tests/test_container_provider.py
```diff
@@ -7,6 +8,16 @@
...
+provider_token = "THE_PROVIDER_TOKEN"
+provider_port = "8443"
+provider_id = 134
+hawkular_hostname = "some-hawkular-hostname.tld"
+hawkular_port = "54321"
```
dkorn: perhaps this can also be a "real" port example, like "443"?

tests/test_container_provider.py
```diff
@@ -40,82 +52,100 @@ class AnsibleModuleFailed(Exception):
 
 @pytest.fixture()
 def miq(miq_api_class, miq_ansible_module, the_provider):
+
+    # TODO(jaryn) This would ideally be inside test cases that need it.
```
dkorn: IINM, this is possible after https://github.com/dkorn/manageiq-ansible-module/pull/15 is merged right?

tests/test_container_provider.py
```diff
@@ -40,82 +52,100 @@ class AnsibleModuleFailed(Exception):
...
     def fail(msg):
         raise AnsibleModuleFailed(msg)
 
     miq_ansible_module.fail_json = fail
-    miq = manageiq.ManageIQ(miq_ansible_module, "http://themanageiq.tld",
+    miq = manageiq.ManageIQ(miq_ansible_module, manageiq_hostname,
                             "The username", "The password")
 
-    miq_api_class.return_value.post.return_value = dict(results=[dict(id=0)])
+    miq_api_class.return_value.post.return_value = dict(results=[
```
dkorn, line 64: can we create another post return value for a delete action? (returns different fields)

dkorn (changes requested):
needs a rebase and few inline comments/questions.

tests/test_container_provider.py
jarovo: The MIQ is using HTTP POST to do delete? Hmm Why doesn't it use DELETE instead?

tests/test_container_provider.py
jarovo: removed

tests/test_container_provider.py
jarovo: done

tests/test_container_provider.py
jarovo: they are now

tests/test_container_provider.py
dkorn: they also support DELETE but IINM when using DELETE you don't get a response (which we need to know that it was successful and track the deletion task).
Do you have an idea how we can support more than one post?

tests/test_container_provider.py
dkorn: :+1:

tests/test_container_provider.py
dkorn: :+1:

tests/test_container_provider.py
jarovo: Interesting.

I think we gonna need another mock for the different post.

 Anyway please merge as I think it is not a topic of this PR

dkorn (approved):

PR #19, opened 2016-09-18
Add provider zone option

dkorn:
It is possible to pass zone name for the added provider. The zone must already exist in manageiq.
If not passed, the 'default' zone will be chosen.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ ManageIQ Python API Client package [manageiq-api-client-python] (https://github.
 The `manageiq_provider` module currently supports adding, updating and deleting an OpenShift provider to manageiq.
 An example playbook `add_provider.yml` is provided and can be run by:
 
-    `$ ansible-playbook add_provider.yml --extra-vars "name=oshift01 type=openshift-origin state=present miq_url=http://localhost:3000 miq_username=user miq_password=****** hostname=oshift01.com port=8443 token=****** metrics=True hawkular_hostname=hawkular01.com hawkular_port=443"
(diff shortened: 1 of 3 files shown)


dkorn:
@jaryn - please take a look at the changes I made to the tests.
@cben, @enoodle, @kwoodson, @simon3z please review

library/manageiq.py
```diff
@@ -34,6 +34,11 @@
       - the openshift provider type
     required: true
     choices: ['openshift-origin', 'openshift-enterprise']
+  zone:
+    description:
+      - the provider zone in manageiq
```
cben: zone name.

library/manageiq.py
```diff
@@ -189,16 +205,17 @@ def add_or_update_provider(self, provider_name, provider_type, endpoints):
             and a short message describing the operation executed
         """
         message = ""
+        zone_id = self.find_zone_by_name(zone or 'default')
```
cben: What if not found and we get zone_id `None`?
- Does update_provider deal with that?
- Does add_new_provider deal with that?

Or do all installations have a `default` zone?

cben (commented):
Have you managed to test this on a real MIQ?

Perhaps add some test where zone is not the same?

library/manageiq.py
dkorn: right, I'll change this

library/manageiq.py
dkorn: IINM all installations have a default zone

dkorn:
@enoodle, @kwoodson, @simon3z any comments?

simon3z:
I gave just a quick look, LGTM

PR #21, opened 2016-09-21
Verify authentications validation after provider addition/update

dkorn:
Add and update provider actions triggers a validation check for the endpoints authentications.
The added functionality verifies that the validation passed and outputs a relevant message.

### DIFF
--- a/library/manageiq_provider.py
+++ b/library/manageiq_provider.py
@@ -362,7 +362,7 @@ class ManageIQ(object):
         else:  # provider doesn't exists, adding it to manageiq
             updates = None
             old_validation_details = {}
-            operation = "add"
+            operation = "addition"
             provider_id = self.add_new_provider(provider_name, provider_type,
                                                 endpoints, zone_id, provider_region)
             roles_with_changes = [e['endpoint']['role'] for e in endpoints]
@@ -372,12 +372,11 @@ class ManageIQ(object):
             if e['endpoint']['role'] in roles_with_changes:
                 authtypes_to_verify.append(e['authentication']['authtype'])
         success, details = self.verify_authenticaion_validation(provider_id, old_validation_details, authtypes_to_verify)
-        operations = {"add": {"past": "added", "present": "addition"},
-                      "update": {"past": "updated", "present": "update"}}
+
         if success:
-            message = "Successfuly {operation} {provider} provider. Authentication: {validation}".format(operation=operations[operation]["past"], provider=provider_name, validation=details)
+            message = "Successful {operation} of {provider} provider. Authentication: {validation}".format(operation=operation, provider=provider_name, validation=details)
         else:
-            message = "Failed to validate provider {provider} after {operation}. Authentication: {validation}".format(operation=operations[operation]["present"], provider=provider_name, validation=details)
+            message = "Failed to validate provider {provider} after {operation}. Authentication: {validation}".format(operation=operation, provider=provider_name, validation=details)
         return dict(
             provider_id=provider_id,
             changed=self.changed,
--- a/library/manageiq_provider.py
+++ b/library/manageiq_provider.py
@@ -1,6 +1,7 @@
 #!/usr/bin/python
 
 import os
+import time
 from ansible.module_utils.basic import *
 from manageiq_client.api import ManageIQClient as MiqApi
 
@@ -150,10 +151,15 @@ class ManageIQ(object):
     """
 
     OPENSHIFT_DEFAULT_PORT = '8443'
+
     provider_types = {
         'openshift-origin': 'ManageIQ::Providers::Openshift::ContainerManager',
         'openshift-enterprise': 'ManageIQ::Providers::OpenshiftEnterprise::ContainerManager',
         'amazon': 'ManageIQ::Providers::Amazon::CloudManager'}
+
+    time_format = '%Y-%m-%dT%H:%M:%SZ'
+    wait_time = 2
+    iterations = 5
 
     def __init__(self, module, url, user, password, verify_ssl, ca_bundle_path):
         self.module        = module
@@ -163,6 +169,28 @@ class ManageIQ(object):
         self.client        = MiqApi(self.api_url, (self.user, self.password), verify_ssl=verify_ssl, ca_bundle_path=ca_bundle_path)
         self.changed       = False
         self.providers_url = self.api_url + '/providers'
+
+    def verify_authenticaion_validation(self, provider_id, update_time):
+        """ Returns a (success, message) tuple:
+                success: True if authentication validation passed, False otherwise
+                message: 'All Valid' if passed, authentication details otherwise
+        """
+        try:
+            epoch_update_time = int(time.mktime(time.strptime(update_time, ManageIQ.time_format)))
+            for i in range(ManageIQ.iterations):
+                result = self.client.get('{providers_url}/{id}/?attributes=authentications'.format(providers_url=self.providers_url, id=provider_id))
+                auths = [{k: auth.get(k, None) for k in ('authtype', 'status', 'status_details', 'updated_on')}
+                         for auth in result['authentications']]
+                # verify that the status reflects the current authentication validation
+                if all(int(time.mktime(time.strptime(auth['updated_on'], ManageIQ.time_format))) > epoch_update_time for auth in auths):
+                    if any(auth['status'] not in ('Valid', None) for auth in auths):
+                        return False, auths
+                    if all(auth['status'] == 'Valid' for auth in auths):
+                        return True, 'All Valid'
+                time.sleep(ManageIQ.wait_time)
+            return False, auths
+        except Exception as e:
+            self.module.fail_json(msg="Failed to get provider data. Error: %s" % e)
 
     def required_updates(self, provider_id, endpoints, zone_id, provider_region):
         """ Checks whether an update is required for the provider
@@ -178,7 +206,7 @@ class ManageIQ(object):
                                 and their values.
         """
         try:
-            result = self.client.get('{providers_url}/{id}/?attributes=authentications,endpoints'.format(providers_url=self.providers_url, id=provider_id))
+            result = self.client.get('{providers_url}/{id}/?attributes=endpoints'.format(providers_url=self.providers_url, id=provider_id))
         except Exception as e:
             self.module.fail_json(msg="Failed to get provider data. Error: {!r}".format(e))
 
@@ -208,12 +236,14 @@ class ManageIQ(object):
         """ Updates the existing provider with new parameters
         """
         try:
+            update_time = time.strftime(ManageIQ.time_format, time.gmtime())
             self.client.post('{api_url}/providers/{id}'.format(api_url=self.api_url, id=provider_id),
                              action='edit',
                              zone={'id': zone_id},
                              connection_configurations=endpoints,
                              provider_region=provider_region)
             self.changed = True
+            return update_time
         except Exception as e:
             self.module.fail_json(msg="Failed to update provider. Error: {!r}".format(e))
 
@@ -221,7 +251,7 @@ class ManageIQ(object):
         """ Adds a provider to manageiq
 
         Returns:
-            the added provider id
+            the added provider id and the creation time
         """
         try:
             result = self.client.post(self.providers_url, name=provider_name,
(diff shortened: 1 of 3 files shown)


dkorn:
@jaryn, @cben, @enoodle, @kwoodson, @yaacov, @simon3z please review

library/manageiq.py
```diff
@@ -113,6 +116,29 @@ def __init__(self, module, url, user, password):
...
         self.providers_url = self.api_url + '/providers'
 
+    def verify_authenticaion_validation(self, provider_id, update_time):
+        """ Returns a (success, message) tuple:
+                success: True if authentication validation passed, False otherwise
+                message: 'All Valid' if passed, authentication details otherwise
+        """
+        try:
+            epoch_update_time = int(time.mktime(time.strptime(update_time, ManageIQ.pattern)))
+            for i in range(5):
```
enoodle: can you make this `5` less hardcoded?

library/manageiq.py
```diff
@@ -113,6 +116,29 @@ def __init__(self, module, url, user, password):
...
+                auths = [{k: auth.get(k, None) for k in ('authtype', 'status', 'status_details', 'updated_on')}
+                         for auth in result['authentications']]
+
+                # verify that the status reflects the current authentication validation
+                if all(int(time.mktime(time.strptime(auth['updated_on'], ManageIQ.pattern))) > epoch_update_time for auth in auths):
+                    if any(auth['status'] not in ('Valid', None) for auth in auths):
+                        return False, auths
+                    if all(auth['status'] == 'Valid' for auth in auths):
+                        return True, 'All Valid'
+                time.sleep(2)
```
enoodle: same goes for this `2` in the sleep.

library/manageiq.py
```diff
@@ -152,28 +178,31 @@ def update_provider(self, provider_id, provider_name, endpoints):
         """ Updates the existing provider with new parameters
         """
         try:
+            update_time = time.strftime(ManageIQ.pattern, time.gmtime())
```
enoodle: Is it possible to get `update_time` from the api response instead of getting it here?

enoodle (changes requested):
Can you get `update_time` the same way that you get `creation_time`?
Other than that, just two nits..

library/manageiq.py
dkorn: Done, thanks!

library/manageiq.py
dkorn: Done, thanks!

library/manageiq.py
dkorn: I tried that, unfortunately, it's not working

cben: do you know if times returned by API are UTC?
will local/API comparisons work if the computers use different time zones?

library/manageiq.py
```diff
@@ -152,28 +178,31 @@ def update_provider(self, provider_id, provider_name, endpoints):
...
+            return update_time
         except Exception as e:
             self.module.fail_json(msg="Failed to update provider. Error: %s" % e)
 
     def add_new_provider(self, provider_name, provider_type, endpoints):
         """ Adds a provider to manageiq
 
         Returns:
-            the added provider id
+            the added provider id and the creation time
```
cben: be clearer it's a tuple eg. "(the added provided id, the creation time)"

library/manageiq.py
```diff
@@ -103,6 +104,9 @@ class ManageIQ(object):
     """
     openshift_provider_types = {'openshift-origin': 'ManageIQ::Providers::Openshift::ContainerManager',
                                 'openshift-enterprise': 'ManageIQ::Providers::OpenshiftEnterprise::ContainerManager'}
+    pattern = '%Y-%m-%dT%H:%M:%SZ'
```
cben: `time_pattern` or `time_format` would be better name IMO.

### Long rant, not a blocker for merging

The hard-coded `Z` bothers me.  It represents UTC, but it's not the only way to represent it.
[MIQ API docs](https://github.com/ManageIQ/manageiq_docs/blob/master/api/overview/json.adoc#basic-types) merely say "Timestamp in ISO8601 format" (and give an example with Z).
Ideally we would be lenient and accept any legal ISO 8601.
_/me reads wiki & stackoverflow..._
Wow, didn't realize [8601](https://en.wikipedia.org/wiki/ISO_8601) is crazy complicated!  Sadly, it's impossible to parse with Python stdlib; there are various external modules, many partially buggy :-(

OK, [RFC 3339](https://tools.ietf.org/html/rfc3339) defines the sane subset used in internet protocols (the format you gave here + optional fractional seconds + mandatory `Z` or `±hh:mm` timezone).
I see API code already has [`normalize_time`](https://github.com/ManageIQ/manageiq/blob/08ff8c0092afc35ae23461923fb52d20520c7562/app/controllers/api/base_controller/normalizer.rb#L66) using ruby `.utc.iso8601` to generate this format, always with 'Z'.  Does it actually run on all time fields?  https://gist.github.com/cben/ee4d703b7c12851bd924879c38b26cd8 => Yes for `created_on`.

=> I'll send a PR to document MIQ API always returns RFC 3339.
    We'll see what people say about committing to `Z`
=> I do recommend accepting optional '.%fZ' fractional seconds, even if AFAICT current MIQ code won't return them — improving timestamp resolution is something that happens to all systems at some point. Requires calling strptime twice I'm afraid (or dependening on external module?).
=> Improved parsing better belongs in the api client lib.
=> A nice quality of strptime is it will raise exception if the format changes, so **I'm fine with merging as for now**  (and I volunteer to follow up if you want).

library/manageiq.py
```diff
@@ -113,6 +117,28 @@ def __init__(self, module, url, user, password):
         self.changed       = False
         self.providers_url = self.api_url + '/providers'
 
+    def verify_authenticaion_validation(self, provider_id, update_time):
+        """ Returns a (success, message) tuple:
```
cben: document params please

library/manageiq.py
```diff
@@ -113,6 +117,28 @@ def __init__(self, module, url, user, password):
...
+                message: 'All Valid' if passed, authentication details otherwise
+        """
+        try:
+            epoch_update_time = int(time.mktime(time.strptime(update_time, ManageIQ.pattern)))
+            for i in range(ManageIQ.iterations):
+                result = self.client.get('{providers_url}/{id}/?attributes=authentications'.format(providers_url=self.providers_url, id=provider_id))
+                auths = [{k: auth.get(k, None) for k in ('authtype', 'status', 'status_details', 'updated_on')}
+                         for auth in result['authentications']]
+                # verify that the status reflects the current authentication validation
+                if all(int(time.mktime(time.strptime(auth['updated_on'], ManageIQ.pattern))) > epoch_update_time for auth in auths):
```
cben: 1. can auth['update_on'] be missing (None)?
2. why `int()`?  the hardcoded format has second resolution but int doesn't add much.
3. why `mktime()`  the struct returned by strptime is perfectly comparable.
4. what if your clock (in update scenario) is slightly ahead of MIQ's clock?  IIUC if auth gets validated quickly enough, it will get stuck on time smaller than `epoch_update_time`, and you'll time out.
5. line is very long.  I'd make helper parse_time() function.  and/or move `for auth in ...` to next line.

library/manageiq.py
cben: Yeah, gmtime returns UTC.  (Unless the computer is misconfigured.)
Would love to somehow check this in tests this but can't think of a meaningful way.
BTW, I'm surprised you don't mock gmtime in tests.  Is update_provider scenario uncovered?

I'd pass around struct_time objects instead of formatting then parsing, but LGTM.

cben: Idea: can you query authentications once before updating/adding, and use time from that as baseline?

cben (commented):
Don't worry, the long rant is not a blocker.  Sorry, I'm a pedant, and timezones always scare me ;-)

library/manageiq.py
cben: Opened https://github.com/ManageIQ/manageiq/issues/11533

dkorn:
> time_pattern or time_format would be better name IMO.

Done, thanks.

library/manageiq.py
dkorn: Done, thanks.

library/manageiq.py
dkorn: Done, thanks.

dkorn:
@cben @enoodle please re-review

library/manageiq_provider.py
```diff
@@ -164,6 +169,48 @@ def __init__(self, module, url, user, password, verify_ssl, ca_bundle_path):
...
+            result = self.client.get('{providers_url}/{id}/?attributes=authentications'.format(providers_url=self.providers_url, id=provider_id))
+            auths = result.get('authentications', [])
+            return {auth['authtype']: auth for auth in auths}
+        except Exception as e:
+            self.module.fail_json(msg="Failed to get provider data. Error: %s" % e)
+
+    def verify_authenticaion_validation(self, provider_id, old_validation_details, authtypes_to_verify):
+        """ Verifies that the provider's authentication validation passed.
+        provider_id            - the provider's id manageiq
+        old_validation_details - a tuple of (last_valid_on, last_invalid_on), representing the last time
```
cben, line 183: This it no longer a tuple, it's prev result of `auths_validation_details()`.

library/manageiq_provider.py
```diff
@@ -164,6 +169,48 @@ def __init__(self, module, url, user, password, verify_ssl, ca_bundle_path):
...
+            details = {t: (new_validation_details[t].get('status', "Validation in progress"),
+                           new_validation_details[t].get('status_details', ''))
+                       for t in authtypes_to_verify}
+            if validations_done:
+                if any(new_validation_details[t]['status'] not in ('Valid', None) for t in authtypes_to_verify):
+                    return False, details
+                if all(new_validation_details[t]['status'] == 'Valid' for t in authtypes_to_verify):
+                    return True, 'All Valid'
+            time.sleep(ManageIQ.WAIT_TIME)
+        return False, details
```
cben, line 212: I think the details returned here might include misleading old `status`, `status_details`.
Should we omit status whose timestamps indicate they're old?  OTOH we're here *because* validation didn't complete yet (or never started for some reason), and we want *some* info on ones who didn't complete.

library/manageiq_provider.py
```diff
@@ -296,35 +343,47 @@ def add_or_update_provider(self, provider_name, provider_type, endpoints, zone,
...
-            )
+            roles_with_changes = [e['endpoint']['role'] for e in endpoints]
+
+        authtypes_to_verify = []
+        for e in endpoints:
+            if e['endpoint']['role'] in roles_with_changes:
+                authtypes_to_verify.append(e['authentication']['authtype'])
+        success, details = self.verify_authenticaion_validation(provider_id, old_validation_details, authtypes_to_verify)
+        operations = {"add": {"past": "added", "present": "addition"},
+                      "update": {"past": "updated", "present": "update"}}
```
cben: Nice, you fixed "updateed" :+1: 
Consider simpler
```
operation, operation_past = "add", "added"
...
operation, operation_past = "update", "update"
```
Or maybe use only "addition"/"update" with "Successful {operation} of {provider} provider." wording.

cben (approved):
Mostly LGTM but I co-wrote this.
@enoodle will you have time to review this?

enoodle (approved):

library/manageiq_provider.py
```diff
@@ -150,11 +151,15 @@ class ManageIQ(object):
     """
 
     OPENSHIFT_DEFAULT_PORT = '8443'
-    provider_types = {
+
+    PROVIDER_TYPES = {
         'openshift-origin': 'ManageIQ::Providers::Openshift::ContainerManager',
         'openshift-enterprise': 'ManageIQ::Providers::OpenshiftEnterprise::ContainerManager',
         'amazon': 'ManageIQ::Providers::Amazon::CloudManager'}
```
yaacov, line 158: kubernetes ?

dkorn: Not yet supported, but definitely a good addition.

@yaacov We welcome contributions, especially on holidays :)

library/manageiq_provider.py
dkorn: Went with the simpler solution (your second one), thank!

PR #23, opened 2016-09-28
Adding ManageIQ Policy Assignment module   

dkorn:
Also added:
* assign_policy.yml example playbook
* test_manageiq_policy_assignment.py unittest
* updated README.md

### DIFF
--- a/assign_policy.yml
+++ b/assign_policy.yml
@@ -1,0 +1,17 @@
+---
+- hosts: hosts
+
+  tasks:
+  - name: Assign policy on an OpenShift provider in ManageIQ
+    manageiq_policy_assignment:
+      entity: '{{ entity }}'
+      entity_name: '{{ entity_name }}'
+      resource: '{{ resource }}'
+      resource_name: '{{ resource_name }}'
+      state: '{{ state }}'
+      miq_url: '{{ miq_url | default(omit) }}'
+      miq_username: '{{ miq_username | default(omit) }}'
+      miq_password: '{{ miq_password | default(omit) }}'
+    register: result
+
+  - debug: var=result
--- a/library/manageiq_policy_assignment.py
+++ b/library/manageiq_policy_assignment.py
@@ -1,0 +1,214 @@
+#!/usr/bin/python
+
+import os
+from ansible.module_utils.basic import *
+from miqclient.api import API as MiqApi
+
+
+DOCUMENTATION = '''
+---
+module: manageiq_policy
+short_description: assign and unassign policies and policy profiles on resources in ManageIQ
+requirements: [ ManageIQ/manageiq-api-client-python ]
+author: Daniel Korn (@dkorn)
+options:
+  miq_url:
+    description:
+      - the manageiq environment url
+    default: MIQ_URL env var if set. otherwise, it is required to pass it
+  miq_username:
+    description:
+      - manageiq username
+    default: MIQ_USERNAME env var if set. otherwise, it is required to pass it
+  miq_password:
+    description:
+      - manageiq password
+    default: MIQ_PASSWORD env var if set. otherwise, it is required to pass it
+  entity:
+    description:
+      - the entity referred to
+    required: true
+    choices: ['policy', 'policy profile']
+  entity_name:
+    description:
+      - the entity name in manageiq
+    required: true
+    default: null
+  resource:
+    description:
+      - the relevant resource type in manageiq
+    required: true
+    choices: ['provider', 'host', 'vm', 'container node', 'pod', 'replicator', 'container image']
+    default: null
+  resource_name:
+    description:
+      - the relevant resource name in manageiq
+    required: true
+    default: null
+  state:
+    description:
+      - On present, it will assign the entity on the resource, if not already assigned
+      - On absent, it will unassign the entity from the resource
+    required: false
+    choices: ['present', 'absent']
+    default: 'present'
+'''
+
+EXAMPLES = '''
+# Assign policy to a provider in ManageIQ
+  manageiq_policy_assignment:
+    entity: 'policy'
+    entity_name: 'example policy'
+    resource_name: 'OpenShift01'
+    resource: 'provider'
+    state: 'present'
+    miq_url: 'http://localhost:3000'
+    miq_username: 'admin'
+    miq_password: '******'
+'''
+
+
+class ManageIQ(object):
+    """ ManageIQ object to execute policy assignments in manageiq
+
+    url      - manageiq environment url
+    user     - the username in manageiq
+    password - the user password in manageiq
+    """
+
+    manageiq_entities = {
+        'policy': 'policies', 'policy profile': 'policy_profiles',
+        'provider': 'providers', 'host': 'hosts', 'vm': 'vms',
+        'container node': 'container_nodes', 'pod': 'container_groups',
+        'replicator': 'container_replicators',
+        'container image': 'container_images'
+    }
+    policy_actions = {
+        'present': 'assign', 'absent': 'unassign'
+    }
+
+    def __init__(self, module, url, user, password):
+        self.module        = module
+        self.api_url       = url + '/api'
+        self.user          = user
+        self.password      = password
+        self.client        = MiqApi(self.api_url, (self.user, self.password))
+        self.changed       = False
+
+    def find_entity_by_name(self, entity_type, entity_name):
+        """ Searches the entity name in ManageIQ.
+
+        Returns:
+            the entity id if it exists in manageiq, None otherwise.
+        """
+        entities_list = getattr(self.client.collections, entity_type)
+        return next((e.id for e in entities_list if e.name == entity_name), None)
+
+    def query_resource_policies_or_profiles(self, entity_type, resource_type, resource_id):
+        """ Returns the policies or policy profiles assigned to the resource.
+        """
+        try:
+            url = '{api_url}/{resource_type}/{resource_id}/{entity_type}?expand=resources'.format(api_url=self.api_url, resource_type=resource_type, resource_id=resource_id, entity_type=entity_type)
+            result = self.client.get(url)
+            return result.get('resources', [])
+        except Exception as e:
+            self.module.fail_json(msg="Failed to query resource {entity_type}: {error}".format(entity_type=entity_type, error=e))
+
+    def entity_assigned(self, entity_type, entity_id, resource_type, resource_id):
+        """Return True if the action is needed on the resource, False otherwise.
+        """
+        assigned_entities = self.query_resource_policies_or_profiles(entity_type, resource_type, resource_id)
+        return any(ae['id'] == entity_id for ae in assigned_entities)
+
+    def execute_action(self, entity_type, entity_id, resource_type, resource_id, action):
+        """Executes the action for the relevant entity on the resource.
+
+        Returns:
+            Whether or not a change took place and a message describing the
+            operation executed.
+        """
+        try:
+            href = '{api_url}/{entity_type}/{entity_id}'.format(api_url=self.api_url, entity_type=entity_type, entity_id=entity_id)
+            url = '{api_url}/{resource_type}/{resource_id}/{entity_type}'.format(api_url=self.api_url, resource_type=resource_type, resource_id=resource_id, entity_type=entity_type)
+            result = self.client.post(url, action=action, resource={'href': href})
+            if result['results'][0]['success']:
+                self.changed = True
+                return dict(
+                    changed=self.changed,
+                    msg=result['results'][0]['message']
+                )
+            else:
+                self.module.fail_json(msg="Failed to {action}: {fail_message}".format(action=action, entity=entity, fail_message=result['results'][0]['message']))
+        except Exception as e:
+            self.module.fail_json(msg="Failed to {action}: {error}".format(action=action, entity=entity, error=e))
+
+    def assign_or_unassign_entity(self, entity, entity_name, resource, resource_name, state):
+        """ Assign or unassign the entity on the manageiq resource.
+
+        Returns:
+            Whether or not a change took place and a message describing the
+            operation executed.
+        """
+        entity_type = self.manageiq_entities[entity]
+        resource_type = self.manageiq_entities[resource]
+        entity_id = self.find_entity_by_name(entity_type, entity_name)
+        if not entity_id:  # entity doesn't exist
+            self.module.fail_json(
+                msg="Failed to {action} {entity}: {entity_name} does not exist in manageiq".format(action=ManageIQ.policy_actions[state], entity=entity, entity_name=entity_name))
+
+        resource_id = self.find_entity_by_name(resource_type, resource_name)
+        if not resource_id:  # resource doesn't exist
+            self.module.fail_json(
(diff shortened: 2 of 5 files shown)


dkorn:
@jaryn, @cben, @enoodle, @kwoodson,@simon3z please review

assign_policy.yml
```diff
@@ -0,0 +1,17 @@
...
+- hosts: hosts
+
+  tasks:
+  - name: Assign policy on an OpenShift provider in ManageIQ
+    manageiq_policy:
+      entity: '{{ entity }}'
+      entity_name: '{{ entity_name }}'
+      resource: '{{ resource }}'
+      resource_name: '{{ resource_name }}'
+      action: '{{ action }}'
```
kwoodson: I think i'd prefer the ansible language here.
action would become state.
present would be assign
absent would be unassign

library/manageiq_policy.py
```diff
@@ -0,0 +1,212 @@
+#!/usr/bin/python
+
+import os
+from ansible.module_utils.basic import *
```
kwoodson: I'd prefer you move this piece to the if **name** section below.  When doing so allows the user to import your module using python and therefore opens up the ability to test, debug, and develop using python.

library/manageiq_policy.py
```diff
@@ -0,0 +1,212 @@
...
+      - the relevant resource type in manageiq
+    required: true
+    choices: ['provider', 'host', 'vm', 'node', 'pod', 'replicator', 'container images']
+    default: null
+  resource_name:
+    description:
+      - the relevant resource name in manageiq
+    required: true
+    default: null
+  action:
```
kwoodson: action => state
`choices = [absent, present]`

library/manageiq_policy.py
```diff
@@ -0,0 +1,212 @@
...
+'''
+
+EXAMPLES = '''
+# Assign policy to a provider in ManageIQ
+  manageiq_policy:
+    entity: 'policy'
+    entity_name: 'example policy'
+    resource_name: 'OpenShift01'
+    resource: 'provider'
+    action: 'assign'
```
kwoodson: state: present

library/manageiq_policy.py
```diff
@@ -0,0 +1,212 @@
...
+    miq_username  = module.params['miq_username']
+    miq_password  = module.params['miq_password']
+    entity        = module.params['entity']
+    entity_name   = module.params['entity_name']
+    resource      = module.params['resource']
+    resource_name = module.params['resource_name']
+    action        = module.params['action']
+
+    manageiq = ManageIQ(module, miq_url, miq_username, miq_password)
+    if action in ['assign', 'unassign']:
```
kwoodson: Not sure if this is needed when the choices are either assign or unassign.

kwoodson (commented):
Consider running pylint on your code.  Could improve indentation, naming, etc.

library/manageiq_policy.py
```diff
@@ -0,0 +1,212 @@
...
+  entity_name:
+    description:
+      - the entity name in manageiq
+    required: true
+    default: null
+  resource:
+    description:
+      - the relevant resource type in manageiq
+    required: true
+    choices: ['provider', 'host', 'vm', 'node', 'pod', 'replicator', 'container images']
```
cben: s/images/image/

library/manageiq_policy.py
```diff
@@ -0,0 +1,212 @@
...
+
+    url      - manageiq environment url
+    user     - the username in manageiq
+    password - the user password in manageiq
+    """
+
+    manageiq_entities = {
+        'policy': 'policies', 'policy profile': 'policy_profiles',
+        'provider': 'providers', 'host': 'hosts', 'vm': 'vms',
+        'node': 'container_nodes', 'pod': 'container_pods',
```
cben: By what criterion does only "container image" have "container " prefix?
In UI I think Container Image and Container Node have "Container " prefix.  (Though that might change in future.)
I don't strongly care what we call them; ideally something predictable & consistent but I don't have concrete solution.

Does `container_pods` actually work with the API?  Shouldn't it be `container_groups`?

library/manageiq_policy.py
```diff
@@ -0,0 +1,212 @@
...
+  resource_name:
+    description:
+      - the relevant resource name in manageiq
+    required: true
+    default: null
+  action:
+    description:
+      - the action used for managing the entity on the resource
+      - On assign, it will assign the entity on the resource, if not already assigned
+      - On absent, it will unassign the entity from the resource
```
cben: s/absent/unassign/ (or s/assign/present/ if you take Kenny's suggestion)

cben:
I find the `entity` vs `resource` interface unclear.
Can you provide a full example assigning a policy to a profile, and a profile to something?
Do you intend this module to later support actions other than assignment — actually creating/editing policies?

dkorn:
> Consider running pylint on your code. Could improve indentation, naming, etc.

@kwoodson you mean in addition to flake8 I'm already running?

assign_policy.yml
dkorn: The thing is, I intend to add more actions in the future, including create and remove policies which IMO fit the 'present' and 'absent' language.
I'm open for suggestions here :)

library/manageiq_policy.py
dkorn: I've seen both forms of imports (on top and in the bottom), I don't really have a strong preference here. @cben what do you think?

```
if __name__ == "__main__":
    main()
```

allows the imports you're reffering to and I'm already testing the modules in `tests/` (should improve of course)

library/manageiq_policy.py
dkorn: explained above https://github.com/dkorn/manageiq-ansible-module/pull/23#discussion_r82587878

library/manageiq_policy.py
dkorn: explained above https://github.com/dkorn/manageiq-ansible-module/pull/23#discussion_r82587878

library/manageiq_policy.py
dkorn: I can drop it for now, added it with thoughts of future actions in mind

library/manageiq_policy.py
dkorn: right, thanks.

dkorn:
> By what criterion does only "container image" have "container " prefix?
> In UI I think Container Image and Container Node have "Container " prefix. (Though that might change in future.)
> I don't strongly care what we call them; ideally something predictable & consistent but I don't have concrete solution.

library/manageiq_policy.py
dkorn: > By what criterion does only "container image" have "container " prefix?

Since assigning policies is currently supported only for _container_ images I didnt want users to confuse them with vm images for example.

> In UI I think Container Image and Container Node have "Container " prefix. (Though that might change in future.)

I'll also add the prefix for nodes. @simon3z any thoughts?  

> Does container_pods actually work with the API? Shouldn't it be container_groups?

it won't work, I'll make the change, thanks!

library/manageiq_policy.py
dkorn: thanks :)

dkorn:
> I find the entity vs resource interface unclear.

entity = policy or policy profile
resource = the relevant resource in manageiq that the policies (and profiles) refer to.
I'm open to other naming suggestions of course :)

> Can you provide a full example assigning a policy to a profile, and a profile to something?

I actually haven't tried that yet, only policies and policy profiles assigned/unassigned on providers.
you can look at the exaple I provided in the documentation of the module or in the README. I can improve it if not clear enough. 

> Do you intend this module to later support actions other than assignment — actually creating/editing policies?

yes, definitely. that's the reason I used `action` instead of the ansible `state`. I'm open for suggestions here as well.

library/manageiq_policy.py
kwoodson: @dkorn,

I was specifically referring to the ansibe import code.  That allows you to import your library in python and work with it.

assign_policy.yml
kwoodson: @dkorn,

I personally do not like the "action" calls.  I'd prefer separate actions move to different ansible modules.  With every action added the code paths increase as well as your parameters.  Then it becomes unclear which parameters are required with which actions.  Separate actions moved into their own modules makes for modular code, loose coupling, and simpler interfaces for the consumers.

dkorn: @kwoodson The policy module is meant to cover all policy and policy profiles related actions:
- assign, unassign, resolve - currently supported
- CRUD actions on policy/policy profile, including actions and conditions - not yet supported in manageiq API

We need to decide with which approach we want to continue here:
a. keep this module only for assign, unassign and resolve (and make the changes you asked with state)
b. add all CRUD missing actions when possible.

@simon3z, @cben, @enoodle what is your opinion?

dkorn:
I've made few changes, the significant ones are:
- separating the policies module into 2:
  manageiq_policy_assignment - this module; supports assign/unassign actions
  manageiq_policy - not yet supported in the API; will include CRUD operations of policies and policy profiles
- using state=present/absent instead of action=assign/unassign, as @kwoodson suggested, now possible since creating the policies is in a separate module.

@cben @kwoodson @twiest @simon3z @enoodle please review

library/manageiq_policy_assignment.py
```diff
@@ -0,0 +1,213 @@
...
+            Whether or not a change took place and a message describing the
+            operation executed.
+        """
+        entity_type = self.manageiq_entities[entity]
+        resource_type = self.manageiq_entities[resource]
+        entity_id = self.find_entity_by_name(entity_type, entity_name)
+        if entity_id:  # entity exists
+            resource_id = self.find_entity_by_name(resource_type, resource_name)
+            if resource_id:  # resource exists
+                assingned = self.entity_assigned(entity_type, entity_id, resource_type, resource_id)
```
twiest: `assingned` should be `assigned`

library/manageiq_policy_assignment.py
```diff
@@ -0,0 +1,213 @@
...
+                    return self.execute_action(entity_type, entity_id, resource_type, resource_id, 'unassign')
+                elif (not assingned) and state == 'present':
+                    return self.execute_action(entity_type, entity_id, resource_type, resource_id, 'assign')
+                else:
+                    return dict(
+                        changed=self.changed,
+                        msg="{entity_name} {entity} already {action}ed".format(entity_name=entity_name, entity=entity, action=ManageIQ.policy_actions[state]))
+            else:
+                self.module.fail_json(
+                    msg="Failed to {action} {entity}: {resource_name} {resource} does not exsit in manageiq".format(action=ManageIQ.policy_actions[state], entity=entity, resource_name=resource_name, resource=resource))
```
twiest: `exsit` should be `exist`

library/manageiq_policy_assignment.py
```diff
@@ -0,0 +1,213 @@
...
+        else:
+            self.module.fail_json(
+                msg="Failed to {action} {entity}: {entity_name} does not exsit in manageiq".format(action=ManageIQ.policy_actions[state], entity=entity, entity_name=entity_name))
```
twiest: same (`exist`)

library/manageiq_policy_assignment.py
```diff
@@ -0,0 +1,213 @@
...
+        """ Aassign or unassign the entity on the manageiq resource.
+
+        Returns:
+            Whether or not a change took place and a message describing the
+            operation executed.
+        """
+        entity_type = self.manageiq_entities[entity]
+        resource_type = self.manageiq_entities[resource]
+        entity_id = self.find_entity_by_name(entity_type, entity_name)
+        if entity_id:  # entity exists
```
twiest: A pattern I like to follow is the "early exit" pattern.

Basically, exit the function as soon as possible instead of having complex if blocks like this. I think it makes the logic much easier to follow.

It will also keep pylint from saying "too many branches" since there are way fewer nested branches.

So on failure checks, have them run as the default case (normally this would be a return, but it's fine in a module to have it be module.fail_json).

So, in this specific case, it'd looks something like this (untested):

``` python

    def assign_or_unassign_entity(self, entity, entity_name, resource, resource_name, state):
        """ Assign or unassign the entity on the manageiq resource.
        Returns:
            Whether or not a change took place and a message describing the
            operation executed.
        """
        entity_type = self.manageiq_entities[entity]
        resource_type = self.manageiq_entities[resource]
        entity_id = self.find_entity_by_name(entity_type, entity_name)

        if not entity_id:  # entity doesn't exist
            self.module.fail_json(
                msg="Failed to {action} {entity}: {entity_name} does not exsit in manageiq".format(action=ManageIQ.policy_actions[state], entity=entity, entity_name=entity_name))


        resource_id = self.find_entity_by_name(resource_type, resource_name)
        if not resource_id:  # resource doesn't exist
            self.module.fail_json(
                msg="Failed to {action} {entity}: {resource_name} {resource} does not exsit in manageiq".format(action=ManageIQ.policy_actions[state], entity=entity, resource_name=resource_name, resource=resource))


        assigned = self.entity_assigned(entity_type, entity_id, resource_type, resource_id)
        if assigned and state == 'absent':
            return self.execute_action(entity_type, entity_id, resource_type, resource_id, 'unassign')


         if (not assigned) and state == 'present':
            return self.execute_action(entity_type, entity_id, resource_type, resource_id, 'assign')


         # Default case is that there's nothing to change
         return dict(
                        changed=self.changed,
                        msg="{entity_name} {entity} already {action}ed".format(entity_name=entity_name, entity=entity, action=ManageIQ.policy_actions[state]))
```

library/manageiq_policy_assignment.py
```diff
@@ -0,0 +1,213 @@
...
+                    changed=self.changed,
+                    msg=result['results'][0]['message']
+                )
+            else:
+                self.module.fail_json(msg="Failed to {action}: {fail_message}".format(action=action, entity=entity, fail_message=result['results'][0]['message']))
+        except Exception as e:
+            self.module.fail_json(msg="Failed to {action}: {error}".format(action=action, entity=entity, error=e))
+
+    def assign_or_unassign_entity(self, entity, entity_name, resource, resource_name, state):
+        """ Aassign or unassign the entity on the manageiq resource.
```
twiest: `Aassign` should be `Assign`

twiest (changes requested):
Looking good, had a few cleanup comments / suggestions.

library/manageiq_policy_assignment.py
dkorn: fixes, thanks!

library/manageiq_policy_assignment.py
dkorn: fixes, thanks!

library/manageiq_policy_assignment.py
dkorn: fixes, thanks!

library/manageiq_policy_assignment.py
dkorn: fixes, thanks!

library/manageiq_policy_assignment.py
dkorn: @twiest I like this suggestion! will make the change, thanks.

dkorn:
@twiest IINM I've made all the changes you asked

twiest (approved):
Looks great!

PR #24, opened 2016-10-06
Renaming several provider options

dkorn:
following @twiest and @cben suggestions (a compromise :))

Renamed:
hostname -> provider_api_hostname
port -> provider_api_port
token -> provider_api_auth_token
    
Also, changing ems API port to default to 8443
@cben @enoodle @simon3z @kwoodson @twiest please review


cben:
LGTM.

Unrelated, but isn't port typically 8443?  Let's make it optional?

dkorn:
> Unrelated, but isn't port typically 8443? Let's make it optional?

@cben if not specified, the default port is indeed 8443. Having said that, it is still mandatory in the UI.

twiest:
Looks great!  :+1:

cben:
@dkorn I don't mind it being required in UI or MIQ API; I meant make it `required=False, default=8443` so you can omit it in ansible playbooks.
Otherwise, we could simply have single `openshift_api_url` — I think the main reason (host, port) pairs are considered more friendly is that having a default port you can frequently just specify host.

dkorn:
> @dkorn I don't mind it being required in UI or MIQ API; I meant make it required=False, default=8443 so you can omit it in ansible playbooks.
> Otherwise, we could simply have single openshift_api_url — I think the main reason (host, port) pairs are considered more friendly is that having a default port you can frequently just specify host.

OK @cben, I'm convinced :)
Will make the change, thanks!

simon3z:
@dkorn would this mean that we'll have specific names for each provider? `openshift_api_hostname`, `ovirt_api_hostname`, etc.? Wouldn't it make sense to aggregate the common ones as we do on the ManageIQ side? E.g. `ems_hostname`, `ems_port`, etc...

dkorn:
> @dkorn would this mean that we'll have specific names for each provider? openshift_api_hostname, ovirt_api_hostname, etc.? Wouldn't it make sense to aggregate the common ones as we do on the ManageIQ side? E.g. ems_hostname, ems_port, etc...

@simon3z great suggestion, I think I'll go with `ems_api_hostname` and `ems_api_port`. @twiest FYI

add_provider.yml
```diff
@@ -10,8 +10,8 @@
       miq_url: '{{ miq_url | default(omit) }}'
       miq_username: '{{ miq_username | default(omit) }}'
       miq_password: '{{ miq_password | default(omit) }}'
-      hostname: '{{ hostname }}'
-      port: '{{ port }}'
+      ems_api_hostname: '{{ ems_api_hostname }}'
+      ems_api_port: '{{ ems_api_port | default(omit) }}'
       token: '{{ token }}'
```
simon3z: @dkorn shouldn't we follow the same naming for these as well? `ems_auth_token`, `ems_hawkular_hostname`, `ems_hawkular_port`, etc.

dkorn: > @dkorn shouldn't we follow the same naming for these as well? ems_auth_token, ems_hawkular_hostname, ems_hawkular_port, etc.

@simon3z I agree, changed. thanks!

dkorn:
@simon3z @cben any other comments or this is ready for merge?

cben:
I'm not sure "ems" = "ExtManagementSystem" means anything to users.
Perhaps s/ems/provider/ ?  Although "provider hostname" doesn't fully make sense.

Anyway LGTM.  I don't think we can be sure what params should be called until we actually support many different provider types.

twiest:
I'm removing my +1 from above. I said the above back when the options still said "openshift". I agree with @cben, ems is meaningless to me. That doesn't help me know what I'm supposed to put as the param to the module.

This also, of course, relates to the ongoing discussion of 1 ansible module per miq provider. "ems" is used because you guys want to create 1 monolithic ansible module, which is IMO not the ansible way (small, cohesive modules) and this leads to a lot of confusion with the params and their names.

simon3z:
> This also, of course, relates to the ongoing discussion of 1 ansible module per miq provider. "ems" is used because you guys want to create 1 monolithic ansible module, which is IMO not the ansible way (small, cohesive modules) and this leads to a lot of confusion with the params and their names.

@twiest the idea of ManageIQ is to have a single-pane of glass for different virtualization/container solutions. Every entity is abstracted: provider, VMs, clusters, etc. Creating a module to handle a specific provider doesn't make sense exactly how it doesn't make sense to create a module to handle a specific VM. If you want to manage a RHEV VM or an OpenStack Instance natively you don't need ManageIQ. In fact on the contrary you use ManageIQ exactly when you want to manage them transparently.

Also, consider that going forward ManageIQ is becoming a platform and the providers are going to be split and maintained separately, which means that we'll see an increasing number of providers being developed and plugged into ManageIQ. Do you expect that each and every provider will come here and contribute an automation which will be 100% the same as the others except for the name of the parameters?

I want this to work out of the box with every provider that the community will produce without having to code anything new for automation (unless a new concept is introduced) [1]

WRT the ansible module we definitely don't want it monolithic, it will be split by entity.

If naming of the specific entries is confusing feel free to suggest one that makes more sense as @cben suggested `provider_api_hostname`, etc.

[1] for that maps as `openshift-origin => ManageIQ::Providers::Openshift::ContainerManager` are problematic.

twiest:
@simon3z This might be better discussed in one of our meetings (there's going to be a lot of back and forth).

But, for now, this is my main argument.

We're talking about trying to fit the miq modules into the _ansible_ way of doing things. Ansible does small, cohesive modules. Look at how many modules there are for ec2, for instance. Or for any other thing for that matter (and these are just in ansible core).

http://docs.ansible.com/ansible/list_of_all_modules.html

What you are describing are internal implementation details of miq, and NOT the conceptual view from an end user's point of view.

To an end user the aws provider and the openshift provider are worlds apart.

The example I gave on the other PR is still apt IMO, which is, if the providers are all the same just different params, then why aren't the providers all configured on the same web ui screen?   IMO, that's because they're conceptually different to the end user.

dkorn:
@cben@simon3z @twiest I've made the changes we discussed, please give me a final OK (or comments if needed)

library/manageiq_provider.py
```diff
@@ -48,19 +48,19 @@
       - the provider zone name in manageiq
     required: false
     default: null
-  hostname:
+  provider_api_hostname:
     description:
-      - the added provider hostname
+      - the added provider hostname API
```
cben: Perhaps "the added provider API hostname" or "hostname of the added provider API" ?

Also since the module can modify existing providers, I'd replace s/added provider/provider/ everywhere.

library/manageiq_provider.py
```diff
@@ -234,8 +235,8 @@ def delete_provider(self, provider_name):
             return dict(task_id=None, changed=self.changed, msg="Provider {provider_name} doesn't exist".format(provider_name=provider_name))
 
     def add_or_update_provider(self, provider_name, provider_type, endpoints, zone):
-        """ Adds an OpenShift containers provider to manageiq or update it's
-        attributes in case a provider with the same name already exists
+        """ Adds a provider to manageiq or update it's  attributes in case
```
cben: s/it's/its/

cben (approved):

library/manageiq_provider.py
dkorn: Done, thanks!

library/manageiq_provider.py
dkorn: I chose: "the provider API hostname" 
and removed "added" from the module docs

library/manageiq_provider.py
```diff
@@ -305,9 +307,9 @@ def main():
     provider_type = module.params['type']
     state         = module.params['state']
     zone          = module.params['zone']
-    hostname      = module.params['hostname']
-    port          = module.params['port']
-    token         = module.params['token']
+    hostname      = module.params['provider_api_hostname']
+    port          = module.params['provider_api_port']
+    token         = module.params['auth_token']
```
twiest: Is the `auth_token` here to auth to the provider?

Assuming so, I think it should probably follow the same pattern as the other two and be cause `provider_api_auth_token`. It's a bit long, but I don't mind that.

twiest (changes requested):
I like `provider_api_hostname` and `provider_api_port`.

`auth_token` feels ambiguous still.

library/manageiq_provider.py
dkorn: I'll make the change

PR #25, opened 2016-10-06
Rename type option to provider_type

dkorn:
The manager_type option specifies the supported provider class names.
This list of supported provider types is documented in the manageiq [REST API docs](http://manageiq.org/docs/reference/latest/api/reference/providers)

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ ManageIQ Python API Client package [manageiq-api-client-python] (https://github.
 The `manageiq_provider` module currently supports adding, updating and deleting an OpenShift provider to manageiq.
 An example playbook `add_provider.yml` is provided and can be run by:
 
-    `$ ansible-playbook add_provider.yml --extra-vars "name=oshift01 type=openshift-origin state=present zone=default miq_url=http://localhost:3000 miq_username=user miq_password=****** provider_api_hostname=oshift01.com provider_api_port=8443 provider_api_auth_token=****** metrics=True hawkular_hostname=hawkular01.com hawkular_port=443"
+    `$ ansible-playbook add_provider.yml --extra-vars "name=oshift01 provider_type=openshift-origin state=present zone=default miq_url=http://localhost:3000 miq_username=user miq_password=****** provider_api_hostname=oshift01.com provider_api_port=8443 provider_api_auth_token=****** metrics=True hawkular_hostname=hawkular01.com hawkular_port=443"
 
 Alternatively, it is possible to add the following environment variables, and remove them from the extra-vars string:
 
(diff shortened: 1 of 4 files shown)


dkorn:
@cben @enoodle @simon3z @kwoodson @jaryn please review

twiest:
@dkorn I really think that each provider in manageIQ should be it's own module. That's really the ansible way (small, cohesive modules that manage 1 thing).

For instance, look at all of the ec2_vpc modules. Those absolutely could have been combined together, but they broke them out into each logical component.

http://docs.ansible.com/ansible/list_of_cloud_modules.html

Is there a reason to have 1 monolithic module for all manageIQ providers?

dkorn:
> @dkorn I really think that each provider in manageIQ should be it's own module. That's really the ansible way (small, cohesive modules that manage 1 thing).
> 
> For instance, look at all of the ec2_vpc modules. Those absolutely could have been combined together, but they broke them out into each logical component.

I agree each module should address a logical component, but IMHO in manageiq a provider is exactly that - a logical concept, to which most actions are shared.

> http://docs.ansible.com/ansible/list_of_cloud_modules.html
> 
> Is there a reason to have 1 monolithic module for all manageIQ providers?

Continuing the point above, all providers support several actions through the API - create, edit, delete, refresh. These actions, along with few other sub-actions (verify authentications, query tasks and more), are almost identical between all providers. They mostly differ in the values passed (same attributes).
Creating a separate module for each provider will cause an unnecessary code duplication. It'll be pretty straightforward when I'll submit the next PR, which support Amazon EC2.

Also, I think that the comparison here isn't fully accurate. IMO, it is like saying that there should be a different module for each storage type in OpenStack (os_volume) or Google cloud (gc_storage).

Let's open it for broader discussion, @cben @enoodle @simon3z what is your opinion?

twiest:
@dkorn ok, I see your point, but I think you're talking about the implementation of the providers and not what they are logically in the system.

As an example, if they're similar, why aren't they configured in the same part of the webui?

They're in different trees because they're logically different in the system (IMO).

I think the main point I'm trying to make is that ManageIQ is presenting them as different entities in the webui, so it's more intuitive to have them as separate modules.

To further my point using your example, in AWS and GCE, the storage is configured on the same page, with different options.

dkorn:
@cben @enoodle @simon3z please share your thoughts

cben:
The UI actually does partial grouping: "Cloud providers" are created in one place but you have to choose specific type (AWS, Azure, OpenStack etc).
The are 3 plausible ansible interfaces:
1. Single `manageiq_provider` module (with either single 'aws' or 2-level 'cloud' -> 'aws' type params).
2. Partially grouped `manageiq_cloud_provider`, `manageiq_container(s)_provider` etc. modules, with type param ('aws', 'openshift').
3. Fully separate `manageiq_aws_provider`, `manageiq_openstack_provider`, `manageiq_openshift_provider` etc. modules.

IMO (2), matching the UI, is worst of both worlds; (1) or (3) are both better.
I think any of the choices is easy enough to take implementation-wise, either way we'll share most but not all logic.  I'm :+1: on merging this PR for now, we could later switch to (3) or (2) with little pain.

**What's more interesting to discuss is what we do with param names — do we strive to make them consistent across types (e.g. `hostname`, `login`, `secret` wherever possible) or do we maximize type-specific meaning (`openshift_api_hostname`,  `access_key_id`, `token` etc)?**
Consistency feels somewhat more natural with (1) and type-specific params somewhat more natural with (3) but both are doable either way.

dkorn:
I updated the PR as we agreed:
*one module for all providers
- short term: Amazon EC2 and Openshift providers will have specific provider type values
- all other providers types are as mentioned in the manageiq REST API docs

@cben @enoodle please review

cben (approved):

PR #26, opened 2016-10-09
Option to add Amazon EC2 Cloud provider to ManageIQ

dkorn:
@cben @enoodle @jaryn please review


library/manageiq_provider.py
```diff
@@ -283,55 +304,66 @@ def main():
             name=dict(required=True),
             zone=dict(required=False, type='str'),
```
cben, line 328: Consider making this `miq_zone` to avoid confusion with AWS zones.

dkorn: I tend to agree, @twiest and @kwoodson what do you think?

library/manageiq_provider.py
```diff
@@ -319,45 +338,56 @@ def main():
...
             ca_bundle_path=dict(required=False, type='str', defualt=None),
+            provider_region=dict(required=False, type='str'),
+            access_key_id=dict(required=False, type='str', no_log=True),
+            secret_access_key=dict(required=False, type='str', no_log=True),
             metrics=dict(required=False, type='bool', default=False),
             hawkular_hostname=dict(required=False),
             hawkular_port=dict(required=False)
         ),
         required_if=[
+            ('manager_type', 'container', ['hostname', 'port', 'token']),
```
cben: `manager_type` appears here and in add_amazon_provider.yml but is nowhere in argument_spec above.  Is it being removed or being added?

library/manageiq_provider.py
```diff
@@ -123,6 +123,19 @@
     provider_api_hostname: 'oshift01.redhat.com'
```
cben, line 123: [unrelated] I wonder if many fields can be omitted in a `state: 'absent'` example?

library/manageiq_provider.py
```diff
@@ -123,6 +123,19 @@
...
     provider_api_port: '8443'
     provider_api_auth_token: '******'
+
+# Add Amazon EC2 Cloud provider to ManageIQ
+  manageiq_provider:
+    name: 'AWS01'
+    provider_type: 'amazon'
+    region: 'us-west-2"
```
cben: s/region/provider_region/?

library/manageiq_provider.py
```diff
@@ -150,14 +164,14 @@ def __init__(self, module, url, user, password, verify_ssl, ca_bundle_path):
         self.changed       = False
         self.providers_url = self.api_url + '/providers'
 
-    def required_updates(self, provider_id, endpoints, zone_id):
+    def required_updates(self, provider_id, endpoints, zone_id, region):
```
cben: I'd also use `miq_zone` and/or `provider_region` or even `aws_region` throught the code to disambiguate (whatever we decide for module params).
Both MIQ and AWS have "zones" and "regions" and it's getting confusing...

library/manageiq_provider.py
```diff
@@ -150,14 +164,14 @@ def __init__(self, module, url, user, password, verify_ssl, ca_bundle_path):
...
         """ Checks whether an update is required for the provider
 
         Returns:
-            Empty Hash (None) - If the hostname and port passed equals the
-                                provider's current values
-            Hash of Changes   - Changes that need to be made if any endpoint is
-                                different than the current values of the
+            Empty Hash (None) - If the hostname, port, zone and region passed equals
```
cben, line 171: [side questions, if these are bugs they're MIQ API bugs]
Q: does changing the MIQ zone of an existing provider actually work?  (do refresh workers relocate to new zone etc)
Q: does changing region of an existing AWS provider work?

library/manageiq_provider.py
```diff
@@ -169,11 +183,12 @@ def required_updates(self, provider_id, endpoints, zone_id):
             self.module.fail_json(msg="Failed to get provider data. Error: %s" % e)
 
         def host_port(endpoint):
-            return {'hostname': endpoint['hostname'], 'port': int(endpoint['port'])}
+            return {'hostname': endpoint.get('hostname') or '', 'port': int(endpoint.get('port') or '0') or ''}
```
cben: `endpoint.get('hostname', '')`.

For the port I'm not sure what you want here; is the purpose of '0' is to become 0 which is "falsy" to trigger `or ''`?
Can we just represent missing port == port 0 ?  `int(endpoint.get('port', '0'))`.

Why do you need these changes now?  Can AWS provider endpoints have missing host/port?

P.S. Turns out it's [actually possible to use port 0](https://daniel.haxx.se/blog/2014/10/25/pretending-port-zero-is-a-normal-one/) in TCP, it's Unix that prevents it.  I don't think we care.
Something like `int(endpoint.get('port', '0'))

library/manageiq_provider.py
```diff
@@ -185,9 +200,11 @@ def host_port(endpoint):
                    if role not in desired_by_role}
         if result['zone_id'] != zone_id:
             updated['zone_id'] = zone_id
+        if provider_region != region:
+            updated['provider_region'] = region
         return {"Updated": updated, "Added": added, "Removed": removed}
 
-    def update_provider(self, provider_id, provider_name, endpoints, zone_id):
+    def update_provider(self, provider_id, provider_name, endpoints, zone_id, region):
```
cben: unused param?  it seems you're not actually updating the region?

library/manageiq_provider.py
```diff
@@ -234,12 +252,13 @@ def find_provider_by_name(self, provider_name):
...
 
-    def generate_endpoint(self, role, hostname, port, authtype, token):
+    def generate_endpoint(self, role, authtype, hostname=None, port=None, token=None, userid=None, password=None):
         """ Returns one provider endpoint hash.
         """
         return {'endpoint': {'role': role, 'hostname': hostname,
                              'port': port},
-                'authentication': {'role': authtype, 'auth_key': token}}
+                'authentication': {'authtype': authtype, 'auth_key': token,
+                                   'userid': userid, 'password': password}}
```
cben, line 262: The 'authentication' dict will contain some Nones, right?  I suppose they become `null` in json.  Is it safe to send irrelevant keys with `null` to the API?

I'm not sure this function is all that helpful now.
Perhaps it should just take the whole authentication dict as one param?

library/manageiq_provider.py
dkorn: which fields did you have in mind?

library/manageiq_provider.py
dkorn: done, thanks!

library/manageiq_provider.py
dkorn: I'll make the `miq_zone` in a separate PR and use `provider_region` throughout the code

library/manageiq_provider.py
dkorn: TBH, regarding both questions, I haven't tried it myself.
changing the MIQ zone should work, otherwise, it's a bug.

library/manageiq_provider.py
dkorn: removed, thanks.

library/manageiq_provider.py
dkorn: we discussed the solution

library/manageiq_provider.py
dkorn: IMO passing the entire authentication isn't the best user experience but we can work on a solution for the endpoint issue in a separate PR

cben:
I did not mean the playbook writer should pass an authentication dict;
I meant the if below where you call generate_endpoints would build the
authentication dict.

It's just a suggestion, your call.

dkorn:
> I did not mean the playbook writer should pass an authentication dict;
> I meant the if below where you call generate_endpoints would build the
> authentication dict.
> 
> It's just a suggestion, your call.

OK, I understand. I'll try to submit it in a separate PR

@cben is there any comment I haven't addressed?

tests/test_container_provider.py
```diff
@@ -47,21 +58,30 @@ def miq_ansible_module():
...
 @pytest.fixture
-def endpoints(miq):
-    endpoints = [
-        miq.generate_endpoint("default", PROVIDER_HOSTNAME, PROVIDER_PORT,
-                              "bearer", PROVIDER_TOKEN),
-        miq.generate_endpoint("hawkular", HAWKULAR_HOSTNAME, HAWKULAR_PORT,
-                              "hawkular", PROVIDER_TOKEN)]
-    yield endpoints
+def openshift_endpoints(miq):
+    openshift_endpoints = [
```
cben: could directly `yield [` here.

cben (approved):

tests/test_container_provider.py
dkorn: Done, thanks.

Merging.

PR #27, opened 2016-10-30
ManageIQ Custom Attributes module

dkorn:
@cben @enoodle @kwoodson @twiest @jaryn @simon3z please review

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,7 @@ ManageIQ Python API Client package [manageiq-api-client-python] (https://github.
     `$ pip install git+https://github.com/ManageIQ/manageiq-api-client-python.git`
 
 
+
 ## Getting Started
 
 ### manageiq_provider module
@@ -46,12 +47,27 @@ Usage:
 To unassign a policy/policy profile on a resource change `state=absent`.
 
 
+### manageiq_custom_attributes module
+
+The `manageiq_custom_attributes` module supports adding, updating and deleting custom attributes on resources in ManageIQ.
+Currently the only resources (entities) that supports custom attributes are vms and providers.
+An example playbook `add_custom_attributes.yml` is provided.
+
+Usage:
+
+    `$ ansible-playbook add_custom_attributes.yml`
+
+To delete a custom attributes change `state=absent`.
+
+
+
 ## SSL Cert Verification
 
 SSL verification for HTTPS requests is enabled by default.
 
 To use a self-signed certificate pass the certificate file or directory path using the ca_bundle_path option: `ca_bundle_path: '/path/to/certfile'`.
 To ignore verifying the SSL certificate pass `verify_ssl: False`
+
 
 
 ## Using Environment Variables
--- a/add_custom_attributes.yml
+++ b/add_custom_attributes.yml
@@ -1,0 +1,20 @@
+---
+- hosts: localhost
+
+  tasks:
+  - name: Add Custom Attributes to Openshift containers provider
+    manageiq_custom_attributes:
+      entity_type: 'provider'
+      entity_name: 'Openshift01'
+      state: 'present'
+      custom_attributes:
+        -  name: 'ca1'
+           value: 'value 1'
+        -  name: 'ca2'
+           value: 'value 2'
+      miq_url: 'https://miq.example.com'
+      miq_username: 'dkorn'
+      miq_password: '******'
+    register: result
+
+  - debug: var=result
--- a/library/manageiq_custom_attributes.py
+++ b/library/manageiq_custom_attributes.py
@@ -1,0 +1,254 @@
+#!/usr/bin/python
+
+import os
+from ansible.module_utils.basic import *
+from miqclient.api import API as MiqApi
+
+
+DOCUMENTATION = '''
+---
+module: manageiq_custom_attributes
+short_description: add, update, delete an entity custom attributes in ManageIQ
+requirements: [ ManageIQ/manageiq-api-client-python ]
+author: Daniel Korn (@dkorn)
+options:
+  miq_url:
+    description:
+      - the manageiq environment url
+    default: MIQ_URL env var if set. otherwise, it is required to pass it
+  miq_username:
+    description:
+      - manageiq username
+    default: MIQ_USERNAME env var if set. otherwise, it is required to pass it
+  miq_password:
+    description:
+      - manageiq password
+    default: MIQ_PASSWORD env var if set. otherwise, it is required to pass it
+  entity_name:
+    description:
+      - the entity name in manageiq to which the custom attributes belongs
+    required: true
+    default: null
+  entity_type:
+    description:
+      - the entity type in manageiq to which the custom attributes belongs
+    required: true
+    default: null
+  state:
+    description:
+      - the state of the custom attributes
+      - On present, it will add the custom attributes to the entity if not
+      already exist, or update the custom attributes if the associated data
+      is different
+      - On absent, it will delete the custom attributes from the entity if exist
+    required: false
+    choices: ['present', 'absent']
+    default: 'present'
+  custom_attributes:
+    description:
+      - the custom attributes of the entity
+    required: true
+    default: null
+'''
+
+EXAMPLES = '''
+# Add custom attributes for Openshift Containers Provider in ManageIQ
+  manageiq_custom_attributes:
+    entity_type: 'provider'
+    entity_name: 'openshift01'
+    state: 'present'
+    miq_url: 'http://localhost:3000'
+    miq_username: 'admin'
+    miq_password: '******'
+    custom_attributes:
+      - name: "ca1"
+        value: "value 1"
+      - name: "ca2"
+        value: "value 2"
+'''
+
+
+class ManageIQCustomAttributes(object):
+    """ ManageIQ object to execute custom attibutes related operations
+    in manageiq
+
+    url      - manageiq environment url
+    user     - the username in manageiq
+    password - the user password in manageiq
+    """
+
+    supported_entities = {'vm': 'vms', 'provider': 'providers'}
+
+    def __init__(self, module, url, user, password):
+        self.module        = module
+        self.api_url       = url + '/api'
+        self.user          = user
+        self.password      = password
+        self.client        = MiqApi(self.api_url, (self.user, self.password))
+        self.changed       = False
+
+    def find_entity_by_name(self, entity_type, entity_name):
+        """ Searches the entity name in ManageIQ.
+
+            Returns:
+                the entity id if it exists in manageiq, None otherwise.
+        """
+        entities_list = getattr(self.client.collections, ManageIQCustomAttributes.supported_entities[entity_type])
+        return next((e.id for e in entities_list if e.name == entity_name), None)
+
+    def get_entity_custom_attributes(self, entity_type, entity_id):
+        """ Returns the entity's custom attributes
+        """
+        try:
+            url = '{api_url}/{entity_type}/{id}?expand=custom_attributes'.format(
+                   api_url=self.api_url,
+                   entity_type=ManageIQCustomAttributes.supported_entities[entity_type],
+                   id=entity_id)
+            result = self.client.get(url)
+            return result.get('custom_attributes', [])
+        except Exception as e:
+            self.module.fail_json(msg="Failed to get {entity_type} custom attributes. Error: {error}".format(
+                entity_type=entity_type, error=e))
+
+    def add_custom_attributes(self, entity_type, entity_id, custom_attributes):
+        """ Returns the added custom attributes """
+        try:
+            url = '{api_url}/{entity_type}/{id}/custom_attributes'.format(
+                    api_url=self.api_url,
+                    entity_type=ManageIQCustomAttributes.supported_entities[entity_type],
+                    id=entity_id)
+            result = self.client.post(url, action='add', resources=custom_attributes)
+            self.changed = True
+            return result['results']
+        except Exception as e:
+            self.module.fail_json(msg="Failed to add the custom attributes. Error: {}".format(e))
+
+    def update_custom_attribute(self, entity_type, entity_id, ca, ca_href):
+        """ Returns the updated custom attributes """
+        try:
+            url = '{api_url}/{entity_type}/{id}/custom_attributes'.format(
+                    api_url=self.api_url,
+                    entity_type=ManageIQCustomAttributes.supported_entities[entity_type],
+                    id=entity_id)
+            ca_object = {'name': ca['name'], 'href': ca_href, 'value': ca['value']}
+            result = self.client.post(url, action='edit', resources=[ca_object])
+            self.changed = True
+            return result['results']
+        except Exception as e:
+            self.module.fail_json(msg="Failed to update the custom attribute {ca_name}. Error: {error}".format(ca_name=ca['name'], error=e))
+
+    def add_or_update_custom_attributes(self, entity_type, entity_name, custom_attributes):
+        """ Adds custom attributes to an entity in manageiq or updates the
+        attributes in case already exists
+
+        Returns:
+            the added or updated custom attributes, whether or not a change
+            took place and a short message describing the operation executed
+        """
+        added, updated = [], []
+        message = ""
+        # check if entity with the type and name passed exists in manageiq
+        entity_id = self.find_entity_by_name(entity_type, entity_name)
+        if not entity_id:  # entity doesn't exist
+            self.module.fail_json(
+                msg="Failed to set the custom attributes. {entity_type} {entity_name} does not exist".format(entity_type=entity_type, entity_name=entity_name))
+
+        entity_cas = self.get_entity_custom_attributes(entity_type, entity_id)
+        for new_ca in custom_attributes:
+            existing_ca = next((ca for ca in entity_cas if ca['name'] == new_ca['name']), None)
+            if existing_ca:
+                if new_ca['value'] != existing_ca['value']:
+                    updated.extend(self.update_custom_attribute(entity_type, entity_id, new_ca, existing_ca['href']))
+            else:
+                added.extend(self.add_custom_attributes(entity_type, entity_id, [new_ca]))
+
+        if added or updated:
+            message = "Successfully set the custom attributes to {entity_name} {entity_type}"
+        else:
+            message = "The custom attributes already exist on {entity_name} {entity_type}"
+
+        return dict(
+            changed=self.changed,
+            msg=message.format(entity_name=entity_name, entity_type=entity_type),
+            updates={"Added": added, "Updated": updated}
+        )
+
+    def delete_custom_attribute(self, ca, ca_href, entity_type, entity_id):
+        """ Returns the deleted custom attribute
+        """
+        try:
+            url = '{api_url}/{entity_type}/{id}/custom_attributes'.format(
+                    api_url=self.api_url,
+                    entity_type=ManageIQCustomAttributes.supported_entities[entity_type],
+                    id=entity_id)
+            ca_object = {'name': ca['name'], 'href': ca_href}
+            result = self.client.post(url, action='delete', resources=[ca_object])
+            self.changed = True
+            return result['results']
+        except Exception as e:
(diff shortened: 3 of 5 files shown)


README.md
```diff
@@ -35,3 +35,13 @@ To update an existing provider pass the changed values together with the require
...
+
+###### manageiq_custom_attributes module
+
+The `manageiq_custom_attributes` module supports adding, updating and deleting custom attributes on resources in ManageIQ.
+Currently the only resources (entities) that supports custom attributes are vms and providers.
+An example playbook `add_custom_attributes.yml` is provided and can be run by:
+
+    `$ ansible-playbook add_custom_attributes.yml`
+
+To delete an custom attributes change `state=absent`.
```
twiest: `an` should be `a`

add_custom_attributes.yml
```diff
@@ -0,0 +1,16 @@
+---
+- hosts: localhost
+
+  tasks:
+  - name: Add Custom Attributes to Openshift containers provider
+    manageiq_custom_attributes:
+      entity_type: 'provider'
+      entity_name: 'Openshift01'
+      state: 'present'
+      custom_attributes: "[{'name': 'ca1', 'value': 'value 1'}, {'name': 'ca2', 'value': 'value 2'}]"
```
twiest: Hmm, can we do this in a more yaml friendly way?

So, something like this:

``` yaml
custom_attributes:
  ca1: value 1
  ca2: value 2
```

library/manageiq_custom_attributes.py
```diff
@@ -0,0 +1,254 @@
...
+        return next((e.id for e in entities_list if e.name == entity_name), None)
+
+    def get_entity_cas(self, entity_type, entity_id):
+        """ Returns the entity's custom attributes
+        """
+        try:
+            url = '{api_url}/{entity_type}/{id}?expand=custom_attributes'.format(
+                   api_url=self.api_url,
+                   entity_type=ManageIQ.supported_entities[entity_type],
+                   id=entity_id)
```
enoodle, line 106: Can this be extracted to an outside function? maybe with an argument to switch the `?expand=` or `/` as the last seperator.

library/manageiq_custom_attributes.py
```diff
@@ -0,0 +1,254 @@
...
+
+        Returns:
+            the added or updated custom attributes, whether or not a change
+            took place and a short message describing the operation executed
+        """
+        added, updated = [], []
+        message = ""
+        # check if entity with the type and name passed exists in manageiq
+        entity_id = self.find_entity_by_name(entity_type, entity_name)
+        if entity_id:  # entity exists
```
twiest: Consider using the "early exit" pattern that I described in your other PR for this entire block.

I think it would make this part of the code much easier to read.

https://github.com/dkorn/manageiq-ansible-module/pull/23/files#r85800999

library/manageiq_custom_attributes.py
```diff
@@ -0,0 +1,254 @@
...
+            ca_list = eval(custom_attributes)
```
twiest: We try to never use `eval`. It's extremely insecure.

I think that this call to `eval` is unnecessary if you change to using a more friendly yaml way to pass this in as Ansible will hand you a dictionary.

enoodle: AFAIK Using `eval` usually introduces many problems and should be avoided. Maybe you can define the `custom_attributes` more precisely to avoid it.

library/manageiq_custom_attributes.py
```diff
@@ -0,0 +1,254 @@
...
+            entity_cas = self.get_entity_cas(entity_type, entity_id)
+            if entity_cas:
+                for new_ca in ca_list:
+                    existing_ca = next((ca for ca in entity_cas if ca['name'] == new_ca['name']), None)
```
enoodle: maybe it should be compared by `section` as well?

library/manageiq_custom_attributes.py
```diff
@@ -0,0 +1,254 @@
...
+                    if existing_ca:
+                        if new_ca['value'] != existing_ca['value']:
+                            updated.extend(self.update_custom_attribute(entity_type, entity_id, new_ca, existing_ca['href']))
+                            message = "Successfuly updated the custom attributes of {entity_name} {entity_type}".format(entity_name=entity_name, entity_type=entity_type)
```
twiest: Since this is in a for loop, isn't message going to get stomped on each loop?

library/manageiq_custom_attributes.py
```diff
@@ -0,0 +1,254 @@
...
+                    else:
+                        added.extend(self.add_custom_attributes(entity_type, entity_id, [new_ca]))
+                        message = "Successfuly added the custom attributes to {entity_name} {entity_type}".format(entity_name=entity_name, entity_type=entity_type)
```
twiest: Same concern (stomped).

library/manageiq_custom_attributes.py
```diff
@@ -0,0 +1,254 @@
...
+            else:  # no custom attributes for this entity
+                added.extend(self.add_custom_attributes(entity_type, entity_id, custom_attributes))
+                message = "Successfuly added the custom attributes to {entity_name} {entity_type}".format(entity_name=entity_name, entity_type=entity_type)
```
twiest: `Successfuly` should be `Successfully`. There are multiple of these throughout the file.

library/manageiq_custom_attributes.py
```diff
@@ -0,0 +1,254 @@
...
+                    entity_type=ManageIQ.supported_entities[entity_type],
+                    id=entity_id)
+            ca_object = {'name': ca['name'], 'href': ca_href}
+            result = self.client.post(url, action='delete', resources=[ca_object])
+            self.changed = True
+            return result['results']
+        except Exception as e:
+            self.module.fail_json(msg="Failed to delete the custom attributes {ca}. Error: {error}".format(ca=ca, error=e))
+
+    def delete_custom_attributes(self, entity_type, entity_name, custom_attributes):
```
enoodle, line 191: This function has a lot of similarities with `add_or_update_custom_attributes`. Its worth the effort to see if it is possible to extract those similarities to a reusable function to simplify them both.

tests/test_custom_attributes.py
```diff
@@ -0,0 +1,162 @@
...
+      "href": "http://miq.example.com/api/providers/266/custom_attributes/6226",
+      "id": 6226,
+      "section": "metadata",
+      "name": "ca1",
+      "value": "value 1",
+      "resource_type": "ExtManagementSystem",
+      "resource_id": 266,
+      "source": "EVM",
+      "serialized_value": "value 1"
+    }]
```
enoodle: I might be wrong because I am not used to pytest but won't it be easier and more clear to compare to `GET_RETURN_VALUES['ca_exist']`?

enoodle (commented):
I like this module. I have some comments from a brief pass on your code. Overall it looks great!

library/manageiq_custom_attributes.py
```diff
@@ -0,0 +1,254 @@
...
+        """ Deletes the custom attributes from the entity, if exist
+
+        Returns:
+            whether or not a change took and a short message including the
+            deleted custom attributes
+        """
+        deleted = []
+        entity_id = self.find_entity_by_name(entity_type, entity_name)
+        if entity_id:
+            ca_list = eval(custom_attributes)
```
twiest: Same concern as above with calling `eval`.

library/manageiq_custom_attributes.py
```diff
@@ -0,0 +1,254 @@
...
+            custom_attributes=dict(required=True, type='str'),
+            miq_url=dict(default=os.environ.get('MIQ_URL', None)),
+            miq_username=dict(default=os.environ.get('MIQ_USERNAME', None)),
+            miq_password=dict(default=os.environ.get('MIQ_PASSWORD', None)),
+        )
+    )
+
+    for arg in ['miq_url', 'miq_username', 'miq_password']:
+        if module.params[arg] in (None, ''):
+            module.fail_json(msg="missing required argument: {}".format(arg))
```
twiest, line 233: To make an argument to ansible required, you can simply add `required=True` to the definition.

So, for miq_url, it'd be this:

``` python
miq_url=dict(required=True, default=os.environ.get('MIQ_URL', None)),
```

If certain params are only required if other params are passed in, you can also use `required_together` like this:

https://github.com/openshift/openshift-tools/blob/b8e2d7659925f89c604c10f5ba2ab43c2402821e/ansible/roles/lib_dyn/library/dyn_record.py#L274-L276

twiest (changes requested):

library/manageiq_custom_attributes.py
cben: is it python const syntax (`ast.literal_eval()`)?  json (`json` module)?
But ansible is natively YAML, instead of taking a string and parsing it, why not have user pass structured data?
[I haven't yet read the PR, this comment might be out of context]

library/manageiq_custom_attributes.py
```diff
@@ -0,0 +1,254 @@
...
+    def find_entity_by_name(self, entity_type, entity_name):
+        """ Searches the entity name in ManageIQ.
+
+            Returns:
+                the entity id if it exists in manageiq, None otherwise.
+        """
+        entities_list = getattr(self.client.collections, ManageIQ.supported_entities[entity_type])
+        return next((e.id for e in entities_list if e.name == entity_name), None)
+
+    def get_entity_cas(self, entity_type, entity_id):
```
cben: I'd call it "get_entity_custom_attributes", or just "get_custom_attributes" especially given that you didn't use "_ca" abbrev in other function names.
(you did in local vars, which is fine)

library/manageiq_custom_attributes.py
dkorn: great suggestion! I definitely agree such function will be useful and not only by this module.
I will create a separate PR to address this (not really in the scope of this change).

library/manageiq_custom_attributes.py
dkorn: I will avoid using the `eval`, thanks

library/manageiq_custom_attributes.py
dkorn: custom attributes are unique by name so it should suffice. TBH, I'm not sure what the section is, only encountered the "metadata" section.

library/manageiq_custom_attributes.py
dkorn: I'll give it a thought, thanks!

tests/test_custom_attributes.py
dkorn: I agree, thanks

README.md
dkorn: done, thanks

add_custom_attributes.yml
dkorn: changed to:

``` yaml
custom_attributes:
  - name: 'ca1'
    value: 'value 1'
  - name: 'ca2'
    value: 'value 2'
```

library/manageiq_custom_attributes.py
dkorn: I agree, will make the change, thanks!

library/manageiq_custom_attributes.py
dkorn: I removed the eval, using "list" type (sequence in yaml)

library/manageiq_custom_attributes.py
dkorn: done, thanks

library/manageiq_custom_attributes.py
dkorn: fixed, thanks

library/manageiq_custom_attributes.py
dkorn: I'm pretty sure it won't work cause required and default are mutually exclusive.

Anyway, this appears in all 3 modules so I'll address it in a separate PR.

dkorn:
@twiest @enoodle @cben IINM I replied or fixed all your comments

library/manageiq_custom_attributes.py
```diff
@@ -0,0 +1,258 @@
...
+    miq_password: '******'
+    custom_attributes:
+      - name: "ca1"
+        value: "value 1"
+      - name: "ca2"
+        value: "value 2"
+'''
+
+
+class ManageIQ(object):
```
enoodle: Question: Why do all the main classes in the modules have the same name? Won't it be better to call it ManageIQCA or some other more specific name?

tests/test_custom_attributes.py
```diff
@@ -0,0 +1,152 @@
...
+
+class AnsibleModuleFailed(Exception):
+    pass
+
+
+@pytest.fixture()
+def miq(miq_api_class, miq_ansible_module, the_provider):
+    # miq_ansible_module.params = {'metrics': True}
+
+    def fail(msg):
```
enoodle, line 92: It will be nice to have failure test, bad arguments / update failure / deletion / CA already exists

enoodle (commented):
Looks really good. I had a question and one suggestion for little more tests.

library/manageiq_custom_attributes.py
dkorn: done, thanks!

tests/test_custom_attributes.py
dkorn: I agree, wanna submit those tests?

library/manageiq_custom_attributes.py
cben: We discussed once making `required` depend on whether the env var exists.  Could have a helper func that returns `{'required': True}` or `{'required': False, default=...}` dict.  :+1: on separate PR.

library/manageiq_custom_attributes.py
```diff
@@ -0,0 +1,259 @@
...
+            the added or updated custom attributes, whether or not a change
+            took place and a short message describing the operation executed
+        """
+        added, updated = [], []
+        message = ""
+        # check if entity with the type and name passed exists in manageiq
+        entity_id = self.find_entity_by_name(entity_type, entity_name)
+        if not entity_id:  # entity doesn't exist
+            self.module.fail_json(
+                msg="Failed to add the custom attributes. {entity_type} {entity_name} does not exist".format(entity_type=entity_type, entity_name=entity_name))
```
cben: You don't know here if you would add or update, perhaps more neutral "Failed to set" ?

library/manageiq_custom_attributes.py
```diff
@@ -0,0 +1,259 @@
...
+        """ Searches the entity name in ManageIQ.
+
+            Returns:
+                the entity id if it exists in manageiq, None otherwise.
+        """
+        entities_list = getattr(self.client.collections, ManageIQCustomAttributes.supported_entities[entity_type])
+        return next((e.id for e in entities_list if e.name == entity_name), None)
+
+    def get_entity_cas(self, entity_type, entity_id):
+        """ Returns the entity's custom attributes
```
cben, line 100: It'd help reader if you add example of what return value looks like.

library/manageiq_custom_attributes.py
```diff
@@ -0,0 +1,259 @@
...
+        """
+        try:
+            url = '{api_url}/{entity_type}/{id}?expand=custom_attributes'.format(
+                   api_url=self.api_url,
+                   entity_type=ManageIQCustomAttributes.supported_entities[entity_type],
+                   id=entity_id)
+            result = self.client.get(url)
+            return result.get('custom_attributes')
```
cben: Can API omit this key despite `?expand=custom_attributes`?
If that "shouldn't happen", I'd write `result['custom_attributes']` and let it explode;
if it simply means there were no attributes, should return [].

library/manageiq_custom_attributes.py
```diff
@@ -0,0 +1,259 @@
...
+        # check if entity with the type and name passed exists in manageiq
+        entity_id = self.find_entity_by_name(entity_type, entity_name)
+        if not entity_id:  # entity doesn't exist
+            self.module.fail_json(
+                msg="Failed to add the custom attributes. {entity_type} {entity_name} does not exist".format(entity_type=entity_type, entity_name=entity_name))
+
+        entity_cas = self.get_entity_cas(entity_type, entity_id)
+        if not entity_cas:  # no custom attributes for this entity
+            added.extend(self.add_custom_attributes(entity_type, entity_id, custom_attributes))
+        else:
```
cben, line 167: I think you can drop this clause if you do `self.get_entity_cas(...) or []` or better yet make did `get_entity_cas()` itself return [].
Then you'd only call `add_custom_attributes()` from single place (which just happens for each of them if none existed).

library/manageiq_custom_attributes.py
```diff
@@ -0,0 +1,259 @@
...
+            for new_ca in custom_attributes:
+                existing_ca = next((ca for ca in entity_cas if ca['name'] == new_ca['name']), None)
```
cben: This _might_ be simpler if you first build a dict keyed by name.  Or not.
Doesn't matter much now; if you intend covering more custom_attributes fields like `source`, `section` later, then would be a better time to decide.

library/manageiq_custom_attributes.py
```diff
@@ -0,0 +1,259 @@
...
+                if existing_ca:
+                    if new_ca['value'] != existing_ca['value']:
+                        updated.extend(self.update_custom_attribute(entity_type, entity_id, new_ca, existing_ca['href']))
+                else:
+                    added.extend(self.add_custom_attributes(entity_type, entity_id, [new_ca]))
+
+        if added:
+            message = "Successfully added the custom attributes to {entity_name} {entity_type}"
+        if updated:
+            message += "Successfully updated the custom attributes of {entity_name} {entity_type}"
```
cben: If both happen, you'd get "..._type}Successfully".  At least add trailing dot.
Or just say "Successfully added/updated" or "Successfully set"; you return detailed `updates` dict anyway.

cben (commented):
Looks great overall!  Small code suggestions.

There are more `custom_attributes` fields you're not covering here (section, source, something about type?...), right?
Does API set them to something fixed?  Do you intend to support them later?

I keep wondering if these modules can have less boilerplate, specifically `delete_custom_attribute` vs `delete_custom_attributes` feel like it shouldn't take _that_ much code... But I don't have concrete suggestions.

dkorn:
> Looks great overall! Small code suggestions.
> 
> There are more custom_attributes fields you're not covering here (section, source, something about type?...), right?

right

> Does API set them to something fixed? Do you intend to support them later?

it does, I definitely intend to add type

> I keep wondering if these modules can have less boilerplate, specifically delete_custom_attribute vs delete_custom_attributes feel like it shouldn't take that much code... But I don't have concrete suggestions.

I'm open for suggestions

library/manageiq_custom_attributes.py
dkorn: I agree, will change to `get_entity_custom_attributes` since you can also get all custom attributes but this method only handles the ones set to the entity.

library/manageiq_custom_attributes.py
dkorn: In what way will it be useful? 
BTW It's the same result from the API request

library/manageiq_custom_attributes.py
dkorn: > Can API omit this key despite ?expand=custom_attributes?

Yes.

> if it simply means there were no attributes, should return [].

you mean instead of returning None? sure, I'll make the change

library/manageiq_custom_attributes.py
dkorn: great suggestion, thanks

library/manageiq_custom_attributes.py
dkorn: I'll go with "set", thanks!

library/manageiq_custom_attributes.py
dkorn: OK, we will discuss it in time, thanks.

library/manageiq_custom_attributes.py
dkorn: great suggestion, I'll make the change.

dkorn:
@cben unless you have other comments I would like to merge this one

cben (approved):

PR #28, opened 2016-10-30
Add SSL Cert Verification support

dkorn:
SSL certificate verification is enabled by default.
To use a self-signed certificate pass the certificate file or directory path using the ca_bundle_path option: `ca_bundle_path: '/path/to/certfile'`.

To ignore verifying the SSL certificate pass `verify_ssl: False`

depends on: https://github.com/ManageIQ/manageiq-api-client-python/pull/10

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,3 +52,9 @@ Alternatively, it is possible to add the following environment variables, and re
     `$ ansible-playbook assign_policy.yml --extra-vars "entity=policy entity_name=openscap resource=provider resource_name=openshift01 state=present"
 
 To unassign a policy/policy profile on a resource change `state=absent`.
+
+### SSL Cert Verification
+
+SSL verification for HTTPS requests is enabled by default.
+To use a self-signed certificate pass the certificate file or directory path using the ca_bundle_path option: `ca_bundle_path: '/path/to/certfile'`.
+To ignore verifying the SSL certificate pass `verify_ssl: False`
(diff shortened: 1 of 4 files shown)


dkorn:
@cben @simon3z @enoodle @twiest @kwoodson please review

cben (approved):
Generally LGTM except it might be unclear what connection this refers to (ansible to MIQ, not MIQ to provider, right?)

dkorn:
> Generally LGTM except it might be unclear what connection this refers to (ansible to MIQ, not MIQ to provider, right?)

@cben right, ansible to MIQ env. If you have a suggestion on how to make it clearer please propose it.

twiest:
The ssl changes look good.

For posterity, as we've discussed before, I still dislike `type: 'openshift-enterprise'`.

dkorn:
> For posterity, as we've discussed before, I still dislike type: 'openshift-enterprise'.

@twiest there's a separate PR that handles the provider type issue https://github.com/dkorn/manageiq-ansible-module/pull/25, still need to change it according to what we've agreed - using the `provider_type` option which will match the types in the ManageIQ API [docs](http://manageiq.org/docs/reference/latest/api/reference/providers). 
As for the other options, as a short term solution, we will support the dedicated amazon and openshift options and will make the changes in the manageiq API to improve user experience.

jarovo:
Could you add 4bf64289790 please?

dkorn:
thanks @jaryn! added your commit.
merging.

PR #30, opened 2016-11-01
Do repr rather than str for the error objects

jarovo:


dkorn:
nice change, thanks!
you need to fix the failed test, rebase (without the merged SSL commit)

perhaps it would be best to use `'{0!r}'.format(...)` instead of `%r` everywhere in the modules for consistency

dkorn:
@jaryn bumping this one up

PR #33, opened 2016-11-16
ManageIQ User Management module

dkorn:
@cben @enoodle @jaryn please review

### DIFF
--- a/create_user.yml
+++ b/create_user.yml
@@ -1,0 +1,18 @@
+---
+- hosts: localhost
+
+  tasks:
+  - name: Create a user in ManageIQ
+    manageiq_user:
+      fullname: 'Daniel Korn'
+      name: 'dkorn'
+      password: '******'
+      group: 'EvmGroup-user'
+      state: 'present'
+      miq_url: 'http://miq.example.com'
+      miq_username: 'admin'
+      miq_password: '******'
+      verify_ssl: False
+    register: result
+
+  - debug: var=result
--- a/library/manageiq_user.py
+++ b/library/manageiq_user.py
@@ -1,0 +1,265 @@
+#!/usr/bin/python
+
+
+DOCUMENTATION = '''
+---
+module: manageiq_user
+short_description: management of users in ManageIQ
+requirements: [ ManageIQ/manageiq-api-client-python ]
+author: Daniel Korn (@dkorn)
+options:
+  miq_url:
+    description:
+      - the manageiq environment url
+    default: MIQ_URL env var if set. otherwise, it is required to pass it
+  miq_username:
+    description:
+      - manageiq username
+    default: MIQ_USERNAME env var if set. otherwise, it is required to pass it
+  miq_password:
+    description:
+      - manageiq password
+    default: MIQ_PASSWORD env var if set. otherwise, it is required to pass it
+  name:
+    description:
+      - the unique userid in manageiq, often mentioned as username
+    required: true
+    default: null
+  fullname:
+    description:
+      - the users' full name
+    required: false
+    default: null
+  password:
+    description:
+      - the users' password
+    required: false
+    default: null
+  group:
+    description:
+      - the name of the group to which the user belongs
+    required: false
+    default: null
+  email:
+    description:
+      - the users' E-mail address
+    required: false
+    default: null
+  state:
+    description:
+      - the state of the user
+      - On present, it will create the user if it does not exist or update the
+      user if the associated data is different
+      - On absent, it will delete the user if it exists
+    required: false
+    choices: ['present', 'absent']
+  verify_ssl:
+    description:
+      - whether SSL certificates should be verified for HTTPS requests
+    required: false
+    default: True
+    choices: ['True', 'False']
+  ca_bundle_path:
+    description:
+      - the path to a CA_BUNDLE file or directory with certificates
+    required: false
+    default: null
+'''
+
+EXAMPLES = '''
+# Create a new user in ManageIQ
+  manageiq_user:
+    name: 'dkorn'
+    fullname: 'Daniel Korn'
+    password: '******'
+    group: 'EvmGroup-user'
+    email: 'dkorn@redhat.com'
+    state: 'present'
+    miq_url: 'http://localhost:3000'
+    miq_username: 'admin'
+    miq_password: '******'
+    verify_ssl: False
+'''
+
+import os
+from miqclient.api import API as MiqApi
+
+
+class ManageIQUser(object):
+    """ ManageIQ object to execute user management operations in manageiq
+
+    url            - manageiq environment url
+    user           - the username in manageiq
+    password       - the user password in manageiq
+    verify_ssl     - whether SSL certificates should be verified for HTTPS requests
+    ca_bundle_path - the path to a CA_BUNDLE file or directory with certificates
+    """
+
+    def __init__(self, module, url, user, password, verify_ssl, ca_bundle_path):
+        self.module        = module
+        self.api_url       = url + '/api'
+        self.user          = user
+        self.password      = password
+        self.client        = MiqApi(self.api_url, (self.user, self.password), verify_ssl=verify_ssl, ca_bundle_path=ca_bundle_path)
+        self.changed       = False
+
+    def find_group_by_name(self, group_name):
+        """ Searches the group name in ManageIQ.
+
+        Returns:
+            the group id if it exists in manageiq, None otherwise.
+        """
+        groups = self.client.collections.groups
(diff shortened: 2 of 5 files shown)


setup.py
```diff
@@ -8,6 +8,6 @@
     url='https://github.com/dkorn/manageiq-ansible-module',
     package_dir={'': 'library'},
     py_modules=["manageiq_provider", "manageiq_policy_assignment",
-                "manageiq_custom_attributes"],
+                "manageiq_custom_attributes", "manageiq_user_management"],
```
kwoodson: I'd prefer to rename to manageiq_user.

library/manageiq_user_management.py
```diff
@@ -0,0 +1,259 @@
...
+            msg="Successfully created the user {userid}: {user_details}".format(userid=userid, user_details=results))
+
+
+def main():
+    module = AnsibleModule(
+        argument_spec=dict(
+            userid=dict(required=True, type='str'),
+            username=dict(required=False, type='str'),
+            password=dict(required=False, type='str', no_log=True),
+            group=dict(required=False, type='str'),
```
kwoodson: As we discussed, multiple groups is probably going to be what we want.  Change to a list.

library/manageiq_user_management.py
```diff
@@ -0,0 +1,259 @@
+#!/usr/bin/python
+
+import os
+from ansible.module_utils.basic import *
```
kwoodson: I like to move the ansible bits under 

`if __name__`

section.  This allows me to import your code in python and work with it.  Consider moving it.

kwoodson (changes requested):
Everything here looks good.  I left a few suggestions and comments.

library/manageiq_user_management.py
dkorn: will do

setup.py
dkorn: @cben @enoodle @simon3z what do you think about the naming of this module

library/manageiq_user_management.py
dkorn: I need to check if that's even possible in manageiq before I do the change

setup.py
simon3z: +1 for `manageiq_user`

dkorn: `manageiq_user` it is :)

library/manageiq_user.py
```diff
@@ -0,0 +1,268 @@
...
+    def delete_user(self, userid):
+        """Deletes the user from manageiq.
+
+        Returns:
+            a short message describing the operation executed.
+        """
+        user_id = self.find_user_by_userid(userid)
+        if not user_id:  # user doesn't exist
+            self.module.fail_json(
+                msg="Failed to delete user {userid}: user {userid} does not exist in manageiq".format(userid=userid))
```
cben: Should this be a failure given `state: absent`?  I think it should simply return changed=False.

library/manageiq_user.py
```diff
@@ -0,0 +1,268 @@
...
+            else:  # updates required
+                results = self.update_user(user_id, userid, username, group_id, password, email)
+                return dict(
+                    changed=self.changed,
+                    msg="Successfully updated the user {userid}: {user_details}".format(userid=userid, user_details=results))
+
+        results = self.create_user(userid, username,  group_id, password, email)
+
+        return dict(
+            changed=self.changed,
```
cben: We've already discussed it, you had good arguments for using `self.changed` (which I don't fully remember), but still it bothers me.
This return with this msg only makes sense knowing that changed==True, while the returns 10 & 5 lines above only make sense knowing that changed==False.

OK, here is a concrete idea: consider moving some of the returns here into methods, e.g. just `return self.update_user(...)` and/or `return self.create_user(..._)`.
I propose the rule of thumb that any method that can `fail_json()` should return a {changed, msg, ...} dict in other cases.

library/manageiq_user.py
dkorn: you're correct, I'll make the change

library/manageiq_user.py
```diff
@@ -0,0 +1,268 @@
...
+      - manageiq password
+    default: MIQ_PASSWORD env var if set. otherwise, it is required to pass it
+  userid:
+    description:
+      - the unique userid in manageiq, often mentioned as username
+    required: true
+    default: null
+  username:
+    description:
+      - the users' full name
```
cben, line 30: So `userid` is "unique id, often mentioned as username" and `username` is "full name"?
AFAICT idempotency here identifies by `userid`, updates `username` if differs.
I think there is a strong tradition the word "username" can only mean the unique identifying string ("dkorn").
There is also some tradition that user "id" is numeric.

=> I suggest renaming this field from `username` to `full_name` (or `user_full_name`), and maybe renaming `userid` to `username`.

cben: On second thoughts, given the context of a "manageiq_user" module, consider `userid` being just `name`.
(Of course assuming the other param is renamed to something like `fullname`)

Many ansible modules use just "name" to identify the thing they operate on; specifically
https://docs.ansible.com/ansible/user_module.html,
https://docs.ansible.com/ansible/win_user_module.html,
https://docs.ansible.com/ansible/postgresql_user_module.html,
https://docs.ansible.com/ansible/os_user_module.html
all use `name` — "Name of the user to create, remove or modify."

@kwoodson what do you think?

dkorn: @cben I agree with `username` / `name` instead of `userid` and `fullname` instead of `username`
@kwoodson is on holiday PTO IMNM, we can always merge it and change it in the future

library/manageiq_user.py
dkorn: @cben I'm OK with going with your proposal, only problem is with `user_update_required` function.
I can move all returns into create/update functions but I'll have to leave that return there.

library/manageiq_user.py
```diff
@@ -0,0 +1,269 @@
...
+            self.module.fail_json(msg="Failed to delete user {userid}: {error}".format(userid=userid, error=e))
+
+    def user_update_required(self, user_id, userid, username, group_id, email):
+        """ Returns true if the username, group id or email passed for the user
+            differ from the user's existing ones, False otherwise.
+        """
+        try:
+            url = "{api_url}/users/{user_id}".format(api_url=self.api_url, user_id=user_id)
+            result = self.client.get(url)
+            return True if result['name'] != username or result['current_group_id'] != group_id or result.get('email') != email else False
```
cben: I'd drop `True if ... else False`, the comparisons give you a boolean.

library/manageiq_user.py
cben: :+1: Rules exist to be broken :-)  It's querying, doesn't mutate manageiq, so reporting "success" is irrelavant.
If you really wanted, you could turn it into `update_user_if_required()` — moving some of the logic from `create_or_update()` into it — but it's also fine as-is.

library/manageiq_user.py
```diff
@@ -0,0 +1,269 @@
...
+        Returns:
+            Whether or not a change took place and a message describing the
+            operation executed.
+        """
+        group_id = self.find_group_by_name(group)
+        if not group_id:  # group doesn't exist
+            self.module.fail_json(
+                msg="Failed to create user {userid}: group {group_name} does not exist in manageiq".format(userid=userid, group_name=group))
+
+        user_id = self.find_user_by_userid(userid)
```
cben, line 208: Heh, this is a confusing line to read.   `userid` is 'fred', `user_id` is 5...

You could start actually using the objects built by `client.collections.foos` instead of passing around ids; but we're still undecided if we want to depend much on the API client, not for this PR...

[no change requested]

dkorn: I agree, this topic deserves its own discussion.

library/manageiq_user.py
dkorn: Done, thanks!

library/manageiq_user.py
dkorn: great suggestion, I went with `update_user_if_required()`.
thanks!

dkorn:
@cben let me know if I can merge this PR from your side

create_user.yml
```diff
@@ -0,0 +1,18 @@
+---
+- hosts: localhost
+
+  tasks:
+  - name: Create a user in ManageIQ
+    manageiq_user:
+      username: 'Daniel Korn'
+      userid: 'dkorn'
```
cben: update to name, fullname

dkorn: just did :)

cben (approved):

PR #35, opened 2016-11-20
SSL cert verification support for policy assignment and custom attributes modules

dkorn:
No new functionality, just adding the SSL cert verifications to modules of which it was missing (were added in parallel to the SSL feature)

### DIFF
--- a/library/manageiq_custom_attributes.py
+++ b/library/manageiq_custom_attributes.py
@@ -49,6 +49,17 @@ options:
       - the custom attributes of the entity
     required: true
     default: null
+  verify_ssl:
+    description:
+      - whether SSL certificates should be verified for HTTPS requests
+    required: false
+    default: True
+    choices: ['True', 'False']
+  ca_bundle_path:
+    description:
+      - the path to a CA_BUNDLE file or directory with certificates
+    required: false
+    default: null
 '''
 
 EXAMPLES = '''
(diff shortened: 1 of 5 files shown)


dkorn:
@enoodle @cben small change, please review

library/manageiq_policy_assignment.py
```diff
@@ -65,6 +76,7 @@
     miq_url: 'http://localhost:3000'
     miq_username: 'admin'
     miq_password: '******'
+    verify_ssl: False
```
cben: this example uses http URL, so this parameter is meaningless?

library/manageiq_custom_attributes.py
```diff
@@ -65,26 +76,29 @@
         value: "value 1"
       - name: "ca2"
         value: "value 2"
+    verify_ssl: False
```
cben: again, meaningless with http: miq_url

cben:
LGTM.  I'd perhaps use true in examples to encourage good practices.  (true is redundant as it's the default but I think we give many redundant fields in examples?)

library/manageiq_policy_assignment.py
dkorn: right, thanks :+1:

library/manageiq_custom_attributes.py
dkorn: right, thanks :+1:

PR #37, opened 2016-11-30
Mind the section of Custom Attributes for their identity

enoodle:
This will use the 'stection' part of the Custom Attribute (If available) to compare two Custom Attributes and decide if they are the same object (for adding/updating/deleting)

@dkorn

### DIFF
--- a/library/manageiq_custom_attributes.py
+++ b/library/manageiq_custom_attributes.py
@@ -150,6 +150,10 @@ class ManageIQCustomAttributes(object):
         except Exception as e:
             self.module.fail_json(msg="Failed to update the custom attribute {ca_name}. Error: {error}".format(ca_name=ca['name'], error=e))
 
+    @staticmethod
+    def compare_custom_attributes(ca1, ca2):
+        return (ca1['name'], ca1['section']) == (ca2['name'], ca2['section'])
+
     def add_or_update_custom_attributes(self, entity_type, entity_name, custom_attributes):
         """ Adds custom attributes to an entity in manageiq or updates the
         attributes in case already exists
@@ -168,7 +172,8 @@ class ManageIQCustomAttributes(object):
 
         entity_cas = self.get_entity_custom_attributes(entity_type, entity_id)
         for new_ca in custom_attributes:
-            existing_ca = next((ca for ca in entity_cas if ca['name'] == new_ca['name']), None)
+            existing_ca = next((ca for ca in entity_cas if
+                self.compare_custom_attributes(ca, new_ca)), None)
             if existing_ca:
                 if new_ca['value'] != existing_ca['value']:
                     updated.extend(self.update_custom_attribute(entity_type, entity_id, new_ca, existing_ca['href']))
@@ -216,7 +221,8 @@ class ManageIQCustomAttributes(object):
 
         entity_cas = self.get_entity_custom_attributes(entity_type, entity_id)
         for new_ca in custom_attributes:
-            ca_href = next((ca['href'] for ca in entity_cas if ca['name'] == new_ca['name']), None)
+            ca_href = next((ca['href'] for ca in entity_cas if
+                self.compare_custom_attributes(ca, new_ca)), None)
             if ca_href:
                 deleted.extend(self.delete_custom_attribute(new_ca, ca_href, entity_type, entity_id))
 
@@ -256,6 +262,10 @@ def main():
     custom_attributes = module.params['custom_attributes']
     verify_ssl        = module.params['verify_ssl']
     ca_bundle_path    = module.params['ca_bundle_path']
+
+    for ca in custom_attributes:
+        if 'section' not in ca:
+            ca['section'] = 'metadata'
 
     manageiq = ManageIQCustomAttributes(module, miq_url, miq_username, miq_password, verify_ssl, ca_bundle_path)
     if state == 'present':
--- a/tests/test_custom_attributes.py
+++ b/tests/test_custom_attributes.py
@@ -15,6 +15,8 @@ PROVIDER_ID = 266
 EXISTING_CA = {"id": 6226, "name": "ca1", "value": "value 1"}
 NEW_CA = {"id": 6227, "name": "ca2", "value": "value 2"}
 UPDATED_CA_VALUE = "new value"
+DEFAULT_SECTION = "metadata"
+DIFFERENT_SECTION = 'section'
 
 
 GET_RETURN_VALUES = {
@@ -23,7 +25,7 @@ GET_RETURN_VALUES = {
         "custom_attributes": [{
             "href": "{miq_hostname}/api/providers/{provider_id}/custom_attributes/{ca_id}".format(miq_hostname=MANAGEIQ_HOSTNAME, provider_id=PROVIDER_ID, ca_id=EXISTING_CA['id']),
             "id": EXISTING_CA['id'],
-            "section": "metadata",
+            "section": DEFAULT_SECTION,
             "name": EXISTING_CA['name'],
             "value": EXISTING_CA['value'],
             "resource_type": "ExtManagementSystem",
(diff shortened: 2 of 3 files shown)


enoodle:
@dkorn I also added a test, PTAL
ping @cben please review

tests/test_custom_attributes.py
```diff
@@ -139,6 +152,22 @@ def test_update_existing_custom_attribute(miq, miq_api_class):
     }
 
 
+def test_compare_custom_attribute_with_section(miq, miq_api_class):
```
dkorn, line 156: The tests checks that a new ca with the same name but a different section will be added right?
If so, can you please add a short docstring to explain it?

library/manageiq_custom_attributes.py
```diff
@@ -150,6 +150,15 @@ def update_custom_attribute(self, entity_type, entity_id, ca, ca_href):
         except Exception as e:
             self.module.fail_json(msg="Failed to update the custom attribute {ca_name}. Error: {error}".format(ca_name=ca['name'], error=e))
 
+    @staticmethod
```
dkorn, line 153: I like the change, but it means that we are assuming that a custom attribute's section cannot be updated (it will add a new one with a different section but same name and value).
If I'm right, are we comfortable with that?
@cben @simon3z

dkorn (changes requested):
one small nit and a general question - are we OK with assuming that a custom attribute's section cannot be updated?

enoodle:
I think that the section is essentially like part of the name. The "effective" name is actually `name + section`.

If we have a custom attribute with a certain name/value/section and we get another one with the same name but different value/section. Should we update the custom attribute or create a new one?
I think we should create a new one with the same name but different section because it is allowed to have two custom attributes with the same name and different sections and if we choose to update we won't be able to have this from this ansible module.

dkorn:
@enoodle I agree with almost everything.
If we get a custom attribute with the same name and same section but different value we **should** update its value.

except that, just add the docstring to the tests and I'll approve the change.

enoodle:
> If we get a custom attribute with the same name and same section but different value we should update its value.

Agreed, this is the situation with this patch - it will compare them as the same Custom Attribute but with different value and will update it. ( https://github.com/dkorn/manageiq-ansible-module/pull/37/files#diff-7c2313b616c8aeac402a9707d03cc668R183 )

I added the docstring to the test

library/manageiq_custom_attributes.py
```diff
@@ -168,7 +177,8 @@ def add_or_update_custom_attributes(self, entity_type, entity_name, custom_attri
 
         entity_cas = self.get_entity_custom_attributes(entity_type, entity_id)
         for new_ca in custom_attributes:
-            existing_ca = next((ca for ca in entity_cas if ca['name'] == new_ca['name']), None)
+            existing_ca = next((ca for ca in entity_cas if
+                ManageIQCustomAttributes.compare_custom_attributes(ca, new_ca)), None)
```
cben: IIRC staticmethod can be called as self.compare_...

dkorn: as a staticmethid I don't mind it to be called from the class (and not from the instance). 
@cben does it matter?

dkorn (approved):

library/manageiq_custom_attributes.py
cben: No, just shorter.  Matters if a subclass would override it, which is irrelevant here.
(`self.` does "virtual" lookup, class name is hard-coded.  
C++ and Java, for unclear reasons, conflated the concept of class methods — that don't receive an instance — with compile-time "static" method lookup.  Python `@staticmethod` while called "static" is actually like `@classmethod` in doing virtual lookup when called from an instance.)

enoodle: I originally made it a `@classmethod` for a reason that I can't even remember. I will change this.

enoodle:
BTW I am aware of the confusing branch name. Its because it took me some time to understand what I really wanted this PR to do. The discussion here clarifies that.

library/manageiq_custom_attributes.py
```diff
@@ -150,6 +150,15 @@ def update_custom_attribute(self, entity_type, entity_id, ca, ca_href):
         except Exception as e:
             self.module.fail_json(msg="Failed to update the custom attribute {ca_name}. Error: {error}".format(ca_name=ca['name'], error=e))
 
+    @staticmethod
+    def compare_custom_attributes(ca1, ca2):
+        if 'section' in ca1 and 'section' in ca2:
```
cben: I'm not sure this logic is right.
If only one of them has a section, the other's section is ignored and we skip to comparing names?

I'd expect something like
```
return (ca1.get('section'), ca1['name']) == (ca2.get('section'), ca2['name'])
```

cben (commented):
- Can this module currently *create* custom attrs with a section?  Do you just add `section` to the hashes under `custom_attributes`?
  The current doc unhelpfully says "      - the custom attributes of the entity".  Can you improve that, and/or include `section` in EXAMPLES?

- What happens if user doesn't specify a section?
  Does API accept that and write CA with section=null in DB?

cben:
I see in #39 that when we use API to create a custom attr without section it defaults to "metadata"?
Is it a documented API behavior?

If so, then we won't be able to update such an attribute, because None != "metadata".
[Well, with current implementation of `compare_custom_attributes` it will maybe work but that's accidental.]

How about always requiring section in playbook?  Explicit is better than implicit IMHO.
Or at least implement "metadata" default on our side, always send section to API, then comparing what we send to what there is becomes easier.

library/manageiq_custom_attributes.py
enoodle: I wanted to only compare with the section if the section is given in both. I thought of being backward compatible and assume the user wanted the default `metadata` or just stay with the current value for section but I see now that it will lead to an unpredictable behavior. I elaborated on that in the main thread comments, But anyway I agree that however we solve this we should compare the tuples like you suggest here.

enoodle:
@cben 
Yes it can create custom attributes with section. I will add it to the examples/documentation, however the section is not mandatory. 
The default value entered by the API to the section field is "metadata". In [1] I made it possible to set it to something else but if nothing is entered then metadata will be placed there (originally it was always `metadata`)
I thought of keeping it simpler and leaving the section field optional for the user (Assuming `metadata` or current value if not present) but as I commented above it will make the behavior unpredictable:
If there are two custom attributes with the same name and different sections and a user of this module will input a custom attribute with that name and no section. Which one should be chosen? both? none? only metadata?
Currently it will catch the "first" one of them and with this patch it will still do that.
I think that to fix it we could either:
1. Force setting the `section` field.
2. Assume in this module as well that the default value should be "metadata" if not set by the user. 


[1]https://github.com/ManageIQ/manageiq/pull/12913

dkorn:
IMO we shouldn't force the user to pass a section, as it isn't required by the API or UI.
creating the `metadata` section in the module before we send the ca is the right approach, option 2 in Erez last comment https://github.com/dkorn/manageiq-ansible-module/pull/37#issuecomment-264513706

enoodle:
@dkorn, @cben I updated to set "metadata" as the default section of custom attributes. I couldn't find how to set it by default from the ansible parameters definitions, so please tell me if there is.

dkorn:
> @dkorn, @cben I updated to set "metadata" as the default section of custom attributes.

good enough

> I couldn't find how to set it by default from the ansible parameters definitions, so please tell me if there is.

I general, you set `default=metadata` in the parameter key inside the `AnsibleModule`  `argument_spec` dict. The problem in this case is that we allow the user to pass a list of custom attributes, without verifying their keys or pass the keys separately. We depend on the API client or server to validate the keys.

dkorn:
LGTM :+1: 

@cben please approve

cben:
LGTM

PR #38, opened 2016-12-01
[WIP] Connected ManageIQ + Openshift containers in one playbook

cben:
Not yet working but I must get it out there to switch to other tasks...

Not sure if this repo is the right place but it's convenient (you need the manageiq ansible module to run this playbook) and perhaps a good example.

Problems/shortcuts:

- [x] ~~The created provider doesn't work.~~  There was a known issue with API not doing the "validate" step, but what I saw here seems worse — the token is not saved anywhere, `Authentication.all` and `AuthToken.all` tables were empty?!

- [ ] The [origin container][1] is supposed to do things on the **host** machine — priviledged, no process isolation, no network isolation!  This has benefits, e.g. your apps will be reachable.
    I failed to make the `bridge` network ManageIQ container talk to a `host` network origin container, so I gave up and switched the origin container to `bridge`.  Now I think it doesn't fully work(?)  And you still can't run 2 of them...

    (I also tried making ManageIQ container `host` network but that conflicted with too many host services e.g. postgres, bad idea...)

- [ ] Currently assumes openshift >= 3.2 / 1.2.

The MIQ container is monolithic and quite heavy on RAM.   In the longer term would be great to use a lower-overhead dev one, and/or separate workers / DB containers to simulate multi-zone/region ManageIQ setups.

I'm not aware of a similarly easy way to run Kubernetes?

[1]: https://docs.openshift.org/latest/getting_started/administrators.html#running-in-a-docker-container


cben:
Hmm test fails on:
```
 ERROR collecting example_miq_with_openshift/openshift-ansible/utils/test/test_utils.py 
ImportError while importing test module '/home/travis/build/dkorn/manageiq-ansible-module/example_miq_with_openshift/openshift-ansible/utils/test/test_utils.py'.
```
I didn't mean it to test code from the `openshift-ansible` submodule, will need to blacklist it...

cben:
Rebased on #42, now provider is created correctly, refresh works :-)

cben:
Closing, I'm not going to work on this here.
Hopefully one day we'll do something similar for integration testing of the modules in Ansible repo.

If anyone has questions on ways to setting up Openshift + connected ManageIQ, ask on https://gitter.im/ManageIQ/manageiq/containers

PR #39, opened 2016-12-01
[WIP] mock MiqAPI to mirror post resources

enoodle:
This will improve the testness of the tests IMO because they will verify better the internal structure of the functions making sure that the data that is sent in the post command has significance as it will be mirrored back to the result value to be asserted.

My solution to the ID numbers is not optimal with more than one custom attribute at a time. I am thinking about adding a logic to parse the `get` data and return the same ID of those custom attributes there if the action is `edit` what do you think?

cc @dkorn , @jaryn , @cben

### DIFF
--- a/tests/test_custom_attributes.py
+++ b/tests/test_custom_attributes.py
@@ -12,8 +12,8 @@ MANAGEIQ_HOSTNAME = "http://miq.example.com"
 PROVIDER_NAME = "openshift01"
 PROVIDER_HOSTNAME = "os01.example.com"
 PROVIDER_ID = 266
-EXISTING_CA = {"id": 6226, "name": "ca1", "value": "value 1"}
-NEW_CA = {"id": 6227, "name": "ca2", "value": "value 2"}
+EXISTING_CA = {"id": 1, "name": "ca1", "value": "value 1"}
+NEW_CA = {"id": 2, "name": "ca2", "value": "value 2"}
 UPDATED_CA_VALUE = "new value"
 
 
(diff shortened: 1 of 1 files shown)


tests/test_custom_attributes.py
```diff
@@ -61,9 +61,38 @@
...
+
+    def post(self, url, action, resources):
+        """
+        Mock the post function to mirror the given resources
+        """
+        posted_cas = []
+        for ca in resources:
+            posted_cas.append({
+                "resource_id": PROVIDER_ID,
+                "section": ca['section'] if 'section' in ca else "metadata",
```
cben, line 82: `ca.get('section', 'metadata')`

tests/test_custom_attributes.py
```diff
@@ -61,9 +61,38 @@
...
 
+class MiqAPIMock(Mock):
+    def __init__(self, *args, **kwargs):
+        super(MiqAPIMock, self).__init__(*args, **kwargs)
+        self.next_id = 1
+
+    def get_new_id(self, action):
+        if action != 'edit':
+            self.next_id = self.next_id + 1
+        return self.next_id
```
cben, line 72: When the action is edit, this returns the last assigned id, which may not be the c.a. you're editing at the moment?  It just happens to work because the update test updates EXISTING_CA which happens to have id 1.
I'd expect `post()` to do the conditional and just preserve id for `edit`.

P.S. python has `+= 1`

cben (commented):
I expect you'd also remove POST_RETURN_VALUES?
Ah, I see you can't.  You kept the assertion the expected `POST_RETURN_VALUES[...]` appears within `result`.

What this improves is (indirectly) checking the posted values, which we previously didn't, right?
Great, but too indirect to my taste.  I'd prefer to see in each test a directly assertion on what we send to API.

I also don't like the indirect way GET/POST_RETURN_VALUES are built from EXISTING/NEW_CA, while still leaving a lot of duplication.

I'm not against merging this, but I'll try to rewrite this more next week.
I think with one helper function I could shorten things, ideally into the tests themselves e.g. update_existing test would look something like:

```
    ...get.return_value = {"custom_attributes": [provider_ca("ca1", "value 1", id=1, href=True)]}
    toupdate = [{'name': "ca1", 'value': 'new value'}]
    api_results = [provider_ca("ca1", "new value", id=1)]
    ...post.return_value = {'results': api_results}
    result = miq.add_or_update_custom_attributes('provider', PROVIDER_NAME, toupdate)
    ...post.assert_called_with(<TODO url>, 'edit', toupdate)
    assert result == {
         'changed': True,
         'msg': ...
         'updates': {
             'Added': [],
             'Updated': api_results
         }
     }
```
Yes I know this goes against the style we already established of having lots of CONSTS but I really find it hard to read.  Extracting common parts is good but I really like seeing the test-specific parts — we had a CA with id=1, we updated id=1, we got back id=1 with new value — inside a test.

cben:
OOPS the test I sketched above didn't check posted value :-/
Edited now to also do `assert_called_with`.

enoodle:
@cben I like your approach of asserting the input values to `post` too. We better have a new PR for that.

cben:
Working on such a rewrite...
Fun find: the current code, under test, does POST to urls like:
```
http://miq.example.com/api/providers/<Mock name='mock.id' id='140074101994888'>/custom_attributes
```
(I think not a bug in the code, just need better mocking...)

PR #42, opened 2016-12-06
Fix endpoint authentication method bug

dkorn:
When passing a provider endpoind to manageiq API its authentication will only save if it includes userid and password or token, but not both (exclusive or).
This behavior is quite unpredictable, as one would expect and error or message when passing both options.

This change separates the generation of openshift provider endpoint from the amazon provider endpoint, thus making sure only one auth method is used for the provider's endpoint.

### DIFF
--- a/library/manageiq_provider.py
+++ b/library/manageiq_provider.py
@@ -253,13 +253,19 @@ class ManageIQ(object):
         providers = self.client.collections.providers
         return next((p.id for p in providers if p.name == provider_name), None)
 
-    def generate_endpoint(self, role, authtype, hostname=None, port=None, token=None, userid=None, password=None):
-        """ Returns one provider endpoint hash.
+    def generate_openshift_endpoint(self, role, authtype, hostname, port, token):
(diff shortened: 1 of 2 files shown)


dkorn:
@cben @enoodle please review

cben (approved):
Can you also add `assert_called_...` to at least one of the amazon tests, checking what we send?

For the record, the code in ManageIQ that made it fail when we also passed `username: null` are:
https://github.com/ManageIQ/manageiq/blob/ebb30f4ad663b4d6f78c5c6bd5d6761c6f73458b/app/models/mixins/authentication_mixin.rb#L134-L136
(`current[:new]` is left unset, exiting on `next if current[:old] == current[:new]` a few lines below)

We'll followup on ManageIQ side to either successfully tolerate extra auth fields with null value, or return an error.

dkorn:
> Can you also add assert_called_... to at least one of the amazon tests, checking what we send?

Done, thanks

cben:
:+1:

Issue #47, opened 2016-12-27
Running update_provider.yml returns an authentication error


pavelzag:
After running an Ansible Playbook to update a provider I am getting the following message:

```
ansible-playbook update_provider_pavel.yml 
 [WARNING]: provided hosts list is empty, only localhost is available


PLAY [localhost] ***************************************************************

TASK [setup] *******************************************************************
ok: [localhost]

TASK [Add Openshift Containers Provider to ManageIQ] ***************************
changed: [localhost]

TASK [debug] *******************************************************************
ok: [localhost] => {
    "result": {
        "changed": true, 
        "msg": "Failed to validate provider Openshift 3 after update. Authentication: {'bearer': (u'Valid', u'Ok')}", 
        "provider_id": 5, 
        "updates": {
            "Added": {}, 
            "Removed": {}, 
            "Updated": {
                "default": {
                    "hostname": "something_different.redhat.com"
                }
            }
        }
    }
}

PLAY RECAP *********************************************************************
localhost                  : ok=3    changed=1    unreachable=0    failed=0   
```

The provider itself was updated in UI

Running the same against an older code doesn't yield such error

cben:
Pavel, can you check what the manageiq UI shows on the provider's page?  Auth status, last refresh status? 

@dkorn I think we timed out waiting for validation? In any case our message here is confusing, "Valid, Ok" doesn't explain what's the problem.

pavelzag:
Hi Beni.
I am getting the following values:


Bearer Authentication	Valid - 8 Minutes Ago
Hawkular Authentication	Valid - 10 Minutes Ago
Last Metrics Collection	Success - 2 Minutes Ago
Last Refresh	
Success - 7 Minutes Ago

cben:
Good, so the module just timed out waiting for auth validation.

We need a better message, and possibly a tunable timeout and maybe not
treating timeout as error, as blocking for over a minute is not always
desirable.
(If validation does complete and fails, that will remain an error.)

Perhaps having more `state` values, like `added` vs `validated` vs
`refreshed` would be good.  Then one could add providers quickly, do other
stuff, then only when needed block for validation/refresh.

I also wonder if API could offer immediate validation instead of queueing
it.

On Dec 28, 2016 8:42 AM, "Pavel Zagalsky" <notifications@github.com> wrote:

Hi Beni.
I am getting the following values:

Bearer Authentication Valid - 8 Minutes Ago
[...]

Issue #44, opened 2016-12-11
Providers with bad API auth tokens can be added successfully


pavelzag:
Tried to run the script with a random token and the action was successful.
Is there a plan to add token validation?

dkorn:
yes, this PR: https://github.com/dkorn/manageiq-ansible-module/pull/21
It was pending for some time and I got back to it a few days ago. I'll revise it today.

cben:
@dkorn I think this can be closed?

dkorn:
@cben definitely, closing.

PR #49, opened 2017-01-04
Refactor auth validation verification method

dkorn:
Also, improves the returned messages in case of failure or time out.
Joint work of Daniel Korn (dkorn@redhat.com) and Beni Cherniavsky-Paskin (cben@redhat.com>)

### DIFF
--- a/library/manageiq_provider.py
+++ b/library/manageiq_provider.py
@@ -184,32 +184,40 @@ class ManageIQ(object):
                                  that the authentication validation occured (success or failure).
         authtypes_to_verify    - a list of autentication types that require validation
 
-        Returns a (success, details) tuple:
-            success: True if authentication validation passed, False otherwise
-            details: 'All Valid' if passed, authentication validation details otherwise
+        Returns a (result, details) tuple:
+            result: 'Valid' if authentication validation passed for all endpoints, 'Invalid' if failed for any endpoint,
+                    'Timed out' if any validation didn't complete in the assigned time
+            details: Authentication validation details, 'Validation didn't complete' in case it timed out
         """
+        def validated(old, new):
+            """ Returns True if the validation timestamp, valid or invalid, is different
+            from the old validation timestamp, False otherwise
+            """
+            return ((old.get('last_valid_on'), old.get('last_invalid_on')) !=
+                    (new.get('last_valid_on'), new.get('last_invalid_on')))
+
         for i in range(ManageIQ.ITERATIONS):
             new_validation_details = self.auths_validation_details(provider_id)
 
-            def validated(old, new):
-                """ Returns True if the validation timestamp, valid or invalid, is different
-                from the old validation timestamp, False otherwise
-                """
-                return ((old.get('last_valid_on'), old.get('last_invalid_on')) !=
-                        (new.get('last_valid_on'), new.get('last_invalid_on')))
+            validations_done = True
+            all_done_valid = "Valid"  # Out of the (re)validated ones.
+            details = {}
+            for t in authtypes_to_verify:
+                old = old_validation_details.get(t, {})
+                new = new_validation_details.get(t, {})
+                if not validated(old, new):
+                    details[t] = "Validation didn't complete"
+                    validations_done = False
+                else:
+                    details[t] = (new.get('status'), new.get('status_details'))
+                    if new.get('status') != 'Valid':
+                        all_done_valid = "Invalid"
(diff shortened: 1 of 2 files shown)


library/manageiq_provider.py
```diff
@@ -381,13 +389,15 @@ def add_or_update_provider(self, provider_name, provider_type, endpoints, zone,
...
-        if success:
+        if result == "Invalid":
+            self.module.fail_json(msg="Failed to Validate provider authentication after {operation}. details: {details}".format(operation=operation, details=details))
+        elif result == "Valid":
             self.refresh_provider(provider_id)
             message = "Successful {operation} of {provider} provider. Authentication: {validation}. Refreshing provider inventory".format(operation=operation, provider=provider_name, validation=details)
-        else:
-            message = "Failed to validate provider {provider} after {operation}. Authentication: {validation}".format(operation=operation, provider=provider_name, validation=details)
+        elif result == "Timed out":
+            message = "provider {provider} Validation after {operation} timed out. Authentication: {validation}".format(operation=operation, provider=provider_name, validation=details)
```
cben: nitpick: Capitalize provider, lowercase "Validation"?

tests/test_manageiq_provider.py
```diff
@@ -304,7 +304,7 @@ def test_will_update_amazon_provider_if_present(miq, miq_api_class, amazon_endpo
         "other region")
     assert res_args == {
         'changed': True,
-        'msg': 'Successful update of {} provider. Authentication: All Valid. Refreshing provider inventory'.format(AMAZON_PROVIDER_NAME),
+        'msg': 'Successful update of {} provider. Authentication: {{}}. Refreshing provider inventory'.format(AMAZON_PROVIDER_NAME),
```
cben, line 307: why "Authentication: {}"?  are `authtypes_to_verify` empty?

cben (commented):
Addresses #47, completes #44.

LGTM except see comment on amazon test, we may have a bug.

@enoodle Can you give impartial opinion on readability of the new logic?

library/manageiq_provider.py
```diff
@@ -184,32 +184,40 @@ def verify_authenticaion_validation(self, provider_id, old_validation_details, a
                                  that the authentication validation occured (success or failure).
         authtypes_to_verify    - a list of autentication types that require validation
 
-        Returns a (success, details) tuple:
-            success: True if authentication validation passed, False otherwise
-            details: 'All Valid' if passed, authentication validation details otherwise
+        Returns a (result details) tuple:
```
enoodle: nit: `(result, details)` missing ','

enoodle (approved):

tests/test_manageiq_provider.py
cben: The field we updated is `provider_region` (stored on ext_management_systems).
I think that will re-connect to another AWS endpoint?  https://docs.aws.amazon.com/general/latest/gr/rande.html
Will/should MIQ re-validate the auth in this case?
Anyway, if it's in 'Updated' dict, shouldn't it have been in `authtypes_to_verify`?

dkorn:
@cben @enoodle @simon3z please review

library/manageiq_provider.py
dkorn: Done, thanks!

library/manageiq_provider.py
dkorn: Done, thanks!

tests/test_manageiq_provider.py
dkorn: @cben yes, `authtypes_to_verify` is empty in this case since the amazon provider region is not part of the default endpoint.
If we want to revalidate after region or zone update we need a change in the current logic:
https://github.com/dkorn/manageiq-ansible-module/blob/93bd719b411042c2469843edc06daf5d4acde056/library/manageiq_provider.py#L390

cben:
:+1: to merge this and address separately if needed.
Can you figure out if MIQ would actually re-validate in such case?

tests/test_manageiq_provider.py
dkorn: @cben I checked with @simon3z, no need to revalidate provider authentications when updating zone or region

dkorn:
> Can you figure out if MIQ would actually re-validate in such case?

I checked with @simon3z, no need to revalidate provider authentications when updating zone or region

PR #52, opened 2017-02-07
Hide ManageIQ Password in Ansible log

dkorn:


dkorn:
@cben @enoodle @ivanhorvath please review

enoodle:
@dkorn this is only for passwords through environment variables? Because I don't see them in the logs when supplying them from the playbook.

dkorn:
> @dkorn this is only for passwords through environment variables? Because I don't see them in the logs when supplying them from the playbook.

it is for all password parameters, not just environment variables. for example `provider_api_auth_token` in `manageiq_provider` module

you can see them in the general ansible log or if you run tasks using `with_items`

cben:
LGTM

PR #55, opened 2017-03-05
Support for provider TLS and custom CA

dkorn:
This change adds the option to verify SSL certificates for a provider's endpoint.
The SSL verification between ManageIQ and the provider is enabled by default, and can be configured by passing `provider_verify_ssl` boolean parameter.
To use a self-signed certificate pass the ca file path option: `provider_ca_path: '/path/to/certfile'`

The change does not add this support in the update functionality since a provider authentication does not re-validate after endpoint update (open issue: https://github.com/ManageIQ/manageiq/issues/13167). Added a `TODO`

### DIFF
--- a/add_openshift_provider.yml
+++ b/add_openshift_provider.yml
@@ -16,8 +16,8 @@
       miq_url: 'http://miq.example.com'
       miq_username: 'admin'
       miq_password: '******'
-      verify_ssl: false
-      provider_verify_ssl
+      miq_verify_ssl: false
+      provider_verify_ssl: false
     register: result
 
   - debug: var=result
--- a/add_openshift_provider.yml
+++ b/add_openshift_provider.yml
@@ -16,7 +16,8 @@
       miq_url: 'http://miq.example.com'
       miq_username: 'admin'
       miq_password: '******'
-      verify_ssl: False
+      verify_ssl: false
+      provider_verify_ssl
     register: result
 
   - debug: var=result
--- a/library/manageiq_provider.py
+++ b/library/manageiq_provider.py
@@ -27,7 +27,7 @@ options:
     default: MIQ_PASSWORD env var if set. otherwise, it is required to pass it
   verify_ssl:
     description:
-      - whether SSL certificates should be verified for HTTPS requests
+      - whether SSL certificates should be verified for HTTPS requests to ManageIQ
     required: false
     default: True
     choices: ['True', 'False']
@@ -75,6 +75,18 @@ options:
       - the provider api auth token
     required: true
     default: null
+  provider_verify_ssl:
+    description:
+      - whether SSL certificates should be verified for HTTPS requests between
+      ManageIQ and the provider
+    required: false
+    default: True
+    choices: ['True', 'False']
+  provider_ca_path:
+    description:
+      - the path to the ca file
+    required: false
+    default: null
   metrics:
     description:
       - whether metrics should be enabled in the provider
@@ -106,7 +118,8 @@ EXAMPLES = '''
     provider_api_hostname: 'oshift01.redhat.com'
     provider_api_port: '8443'
     provider_api_auth_token: '******'
-    verify_ssl: False
+    verify_ssl: false
+    provider_verify_ssl: false
     metrics: True
     hawkular_hostname: 'hawkular01.redhat.com'
     hawkular_port: '443'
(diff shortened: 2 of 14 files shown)


dkorn:
@cben @simon3z @enoodle please review

add_openshift_provider.yml
```diff
@@ -16,7 +16,8 @@
       miq_url: 'http://miq.example.com'
       miq_username: 'admin'
       miq_password: '******'
-      verify_ssl: False
+      verify_ssl: false
+      provider_verify_ssl
```
cben: syntax error

library/manageiq_provider.py
```diff
@@ -75,6 +75,18 @@
       - the provider api auth token
     required: true
     default: null
+  provider_verify_ssl:
+    description:
+      - whether SSL certificates should be verified for HTTPS requests between
+      ManageIQ and the provider
```
cben, line 81: bad indentation, add 2 spaces to continue the bullet item.
I'm not sure what's up with ansible wanting sentence per bullet [https://github.com/ansible/ansible/blob/devel/examples/DOCUMENTATION.yml] but whatever...

You can test the doc parses with `ansible-doc -M ./library/ manageiq_provider -vvv`, which has completely useless error messages.  I now tried and figured out your modules need a top-level `description:` in addition to `short_descriptions:`, AND to be valid yaml (a few indentation issues like this), and then it's happy.

library/manageiq_provider.py
```diff
@@ -75,6 +75,18 @@
...
+    required: false
```
cben, line 82: tip: you can generally omit `required: false` (it's the default), especially if you also specify a `default:`.
https://docs.ansible.com/ansible/dev_guide/developing_modules_documenting.html#documentation-block

library/manageiq_provider.py
```diff
@@ -333,11 +349,18 @@ def find_provider_by_name(self, provider_name):
         providers = self.client.collections.providers
         return next((p.id for p in providers if p.name == provider_name), None)
 
-    def generate_auth_key_config(self, role, authtype, hostname, port, token):
+    def generate_auth_key_config(self, role, authtype, hostname, port, token, provider_verify_ssl, provider_ca_path):
         """ Returns an openshift provider endpoint dictionary.
         """
+        provider_ca_content = ''
```
cben: Make the default None.
Empty string will be treated the same, but null/nil is semantically cleaner representation of "don't use a custom CA".

tests/test_manageiq_provider.py
```diff
@@ -14,6 +14,8 @@
 PROVIDER_HOSTNAME = "some-provider-hostname.tld"
 PROVIDER_TOKEN = "THE_PROVIDER_TOKEN"
 PROVIDER_PORT = 8443
+PROVIDER_VERIFY_SSL = False
+PROVIDER_CA_PATH = ""
```
cben, line 18: can you make a less trivial test?

[consider moving file reading into main() so you can test without creating temp files]

cben (commented):
Thanks for the addition.  Mostly LGTM.

I don't want to block on https://github.com/ManageIQ/manageiq/issues/13167.
Can you also add update code?  What will happen?
It would poll for re-validation and time out?  I think that's NOTABUG from ansible side.

TBD later (as discussed offline):
- I'm still thinking whether we should also send `security_protocol`, and whether user should specify it or we should fill it.  Let's not block on that, for now this is usable.
- I'm afraid we'll need separate hawkular options (defaulting from provider_{verify_ssl,ca_path} if possible).

library/manageiq_provider.py
dkorn: thanks

library/manageiq_provider.py
dkorn: I think I'll keep it, for now at least, to maintain consistency and give a little more info to the users.

library/manageiq_provider.py
dkorn: I opened https://github.com/dkorn/manageiq-ansible-module/issues/56
This should be addressed in a separate PR IMO

add_openshift_provider.yml
dkorn: done, thanks

dkorn:
> I don't want to block on ManageIQ/manageiq#13167.
> Can you also add update code? What will happen?
> It would poll for re-validation and time out? I think that's NOTABUG from ansible side.

I prefer to include this update change in a separate PR, since the change in manageiq affects the entire use of containers providers in modules.
Meanwhile, let's keep the TODO and open an issue.

cben (approved):

cben:
You need to update test to expect sending `'certificate_authority': None` rather than `''`.

dkorn:
> You need to update test to expect sending 'certificate_authority': None rather than ''.

Done, thanks!

PR #57, opened 2017-03-16
Tag assignment module

dkorn:
See `assign_tag.yml` playbook for tag assignment example.
Also added tests.

### DIFF
--- a/library/manageiq_tag_assignment.py
+++ b/library/manageiq_tag_assignment.py
@@ -1,0 +1,229 @@
+#!/usr/bin/python
+
+import os
+from ansible.module_utils.basic import *
+from manageiq_client.api import ManageIQClient as MiqApi
+
+
+DOCUMENTATION = '''
+---
+module: manageiq_tag_assignment
+short_description: assign and unassign tags on resources in ManageIQ
+requirements: [ ManageIQ/manageiq-api-client-python ]
+author: Daniel Korn (@dkorn)
+options:
+  miq_url:
+    description:
+      - the manageiq environment url
+    default: MIQ_URL env var if set. otherwise, it is required to pass it
+  miq_username:
+    description:
+      - manageiq username
+    default: MIQ_USERNAME env var if set. otherwise, it is required to pass it
+  miq_password:
+    description:
+      - manageiq password
+    default: MIQ_PASSWORD env var if set. otherwise, it is required to pass it
+  tags:
+    description:
+      - list of dictionaries representing tags, each includes 'name' (tag name) and 'category' (category name) keys
+      - categories have a 'single_value': True/False. assigning a tag in a single_value: True category
+        clears other tags from that category.
+    required: true
+    default: null
+  resource:
+    description:
+      - the relevant resource type in manageiq
+    required: true
+    choices: ['provider', 'host', 'vm', 'blueprint', 'category', 'cluster', 'data store', 'group', 'resource pool', 'service', 'service template', 'template', 'tenant', 'user']
+    default: null
+  resource_name:
+    description:
+      - the relevant resource name in manageiq
+    required: true
+    default: null
+  state:
+    description:
+      - On present, it will assign the tag on the resource, if not already assigned
+      - On absent, it will unassign the tag from the resource
+    required: true
+    choices: ['present', 'absent']
+    default: null
+  miq_verify_ssl:
+    description:
+      - whether SSL certificates should be verified for HTTPS requests
+    required: false
+    default: True
+    choices: ['True', 'False']
+  ca_bundle_path:
+    description:
+      - the path to a CA_BUNDLE file or directory with certificates
+    required: false
+    default: null
+'''
+
+EXAMPLES = '''
+# Assign tag on a provider in ManageIQ
+  manageiq_tag_assignment:
+    tags:
+    - category: environment
+      name: prod
+    - category: owner
+      name: prod_ops
+    resource_name: 'OpenShift01'
+    resource: 'provider'
+    state: 'present'
+    miq_url: 'http://localhost:3000'
+    miq_username: 'admin'
+    miq_password: '******'
+'''
+
+
+class ManageIQTagAssignment(object):
+    """ ManageIQ object to execute tag assignments in manageiq
+
+    url            - manageiq environment url
+    user           - the username in manageiq
+    password       - the user password in manageiq
+    miq_verify_ssl - whether SSL certificates should be verified for HTTPS requests
+    ca_bundle_path - the path to a CA_BUNDLE file or directory with certificates
+    """
+
+    manageiq_entities = {
+        'provider': 'providers', 'host': 'hosts', 'vm': 'vms',
+        'category': 'categories', 'cluster': 'clusters', 'data store': 'data_stores',
+        'group': 'groups', 'resource pool': 'resource_pools', 'service': 'services',
+        'service template': 'service_templates', 'template': 'templates',
+        'tenant': 'tenants', 'user': 'users', 'blueprint': 'blueprints'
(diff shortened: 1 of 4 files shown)


dkorn:
@cben @enoodle @simon3z please review

assign_tag.yml
```diff
@@ -0,0 +1,21 @@
...
+      state: present
+      miq_url: http://miq.example.com
+      miq_username: admin
+      miq_password: ******
+      miq_verify_ssl: false
+      tags:
+      - name: prod
+        category: environment
+      - name: prod_ops
+        category: owner
```
cben: nitpick: put name after category

library/manageiq_tag_assignment.py
```diff
@@ -0,0 +1,224 @@
...
+      - manageiq username
+    default: MIQ_USERNAME env var if set. otherwise, it is required to pass it
+  miq_password:
+    description:
+      - manageiq password
+    default: MIQ_PASSWORD env var if set. otherwise, it is required to pass it
+  tags:
+    description:
+      - list of dictionaries representing tags, each includes the tag name and category name
+      - assigning tags with the same category is not allowed in ManageIQ
```
cben: not quite true.  categories have a single_value: true/false.  assigning one tag in a single_value: true category clears other tag(s) from that category.

assign_tag.yml
dkorn: done, thanks

library/manageiq_tag_assignment.py
dkorn: Interesting didn't know it.
Will change the documentation, thanks!

library/manageiq_tag_assignment.py
```diff
@@ -0,0 +1,225 @@
...
+    ca_bundle_path - the path to a CA_BUNDLE file or directory with certificates
+    """
+
+    manageiq_entities = {
+        'provider': 'providers', 'host': 'hosts', 'vm': 'vms',
+        'category': 'categories', 'cluster': 'clusters', 'data store': 'data_stores',
+        'group': 'groups', 'resource pool': 'resource_pools', 'service': 'services',
+        'service template': 'service_templates', 'template': 'templates',
+        'tenant': 'tenants', 'user': 'users', 'blueprint': 'blueprints'
+    }
```
cben, line 98: We already have these somewhere (policy assignment?).  Can you move it to a shared module?

library/manageiq_tag_assignment.py
```diff
@@ -0,0 +1,225 @@
...
+    description:
+      - manageiq username
+    default: MIQ_USERNAME env var if set. otherwise, it is required to pass it
+  miq_password:
+    description:
+      - manageiq password
+    default: MIQ_PASSWORD env var if set. otherwise, it is required to pass it
+  tags:
+    description:
+      - list of dictionaries representing tags, each includes the tag name and category name
```
cben: can you word it so it's clear you expect `name` and `category` keys?

library/manageiq_tag_assignment.py
dkorn: Though several entities are shared it's not the same dict as the one in the policy assignment module.

Some other functionalities are shared across several modules but we still haven't discussed the best solution for modules code sharing.

library/manageiq_tag_assignment.py
dkorn: done, thanks :)

library/manageiq_tag_assignment.py
```diff
@@ -0,0 +1,225 @@
...
+
+    def execute_action(self, resource_type, resource_id, tags, action):
+        """Executes the action for the resource tag
+        """
+        url = '{api_url}/{resource_type}/{resource_id}/tags'.format(api_url=self.api_url, resource_type=resource_type, resource_id=resource_id)
+        try:
+            result = self.client.post(url, action=action, resources=tags)
+        except Exception as e:
+            self.module.fail_json(msg="Failed to {action} tag: {error}".format(action=action, error=e))
+        for result in result['results']:
```
cben: Replacing same var with its parts, confusing.  `for res in result` or something.  Or maybe rename what you get from client.post `response`.

library/manageiq_tag_assignment.py
```diff
@@ -0,0 +1,225 @@
...
+            if result['success']:
+                self.changed = True
+            else:
+                self.module.fail_json(msg="Failed to {action}: {fail_message}".format(action=action, entity=entity, fail_message=result['results'][0]['message']))
```
cben: I guess `result['results'][0]` here should be simply `result` or however you rename the `for` loop variable?

library/manageiq_tag_assignment.py
```diff
@@ -0,0 +1,225 @@
...
+        resource_id = self.find_entity_by_name(resource_type, resource_name)
+        if not resource_id:  # resource doesn't exist
+            self.module.fail_json(
+                msg="Failed to {action} tag: {resource_name} {resource} does not exist in manageiq".format(action=ManageIQ.actions[state], resource_name=resource_name, resource=resource))
+
+        tags_to_execute = []
+        for tag in tags:
+            tag_name = tag['name']
+            category_name = tag['category']
+            assigned = self.is_tag_assigned(resource_type, resource_id, category_name, tag_name)
```
cben: This queries resource's all tags, looks for one among them, repeats...
Can you query once, build a set, and do set difference / intersection?

library/manageiq_tag_assignment.py
```diff
@@ -0,0 +1,225 @@
...
+    resource_name: 'OpenShift01'
+    resource: 'provider'
+    state: 'present'
+    miq_url: 'http://localhost:3000'
+    miq_username: 'admin'
+    miq_password: '******'
+'''
+
+
+class ManageIQ(object):
```
cben: You usually name these classes according to module name, no?  ManageIQTagAssignment?

dkorn: right, thanks!

library/manageiq_tag_assignment.py
dkorn: I'll go with `response` and leave `result` in the for loop:
```python
for result in response['results']
```

thanks!

library/manageiq_tag_assignment.py
dkorn: done, thanks

library/manageiq_tag_assignment.py
dkorn: Great suggestion, done!
Can you take a look? @cben

library/manageiq_tag_assignment.py
```diff
@@ -0,0 +1,229 @@
...
+        try:
+            url = '{api_url}/{resource_type}/{resource_id}/tags?expand=resources'.format(api_url=self.api_url, resource_type=resource_type, resource_id=resource_id)
+            response = self.client.get(url)
+        except Exception as e:
+            self.module.fail_json(msg="Failed to query {resource_type} tags: {error}".format(resource_type=resource_type, error=e))
+        tags = response.get('resources', [])
+        tags_set = set([tag['name'] for tag in tags])
+        return tags_set
+
+    def is_tag_assigned(self, resource_type, resource_id, category_name, tag_name, assigned_tags):
```
cben: You can drop resource_type, resource_id params now.
Perhaps even reduce this to a `full_tag_name(category_name, tag_name)` or `full_tag_name(tag_dict)` function, so caller would check `full_tag_name(...) in assigned_tags`?

LGTM otherwise.

dkorn: @cben another great suggestion, revised in a new commit

cben (approved):

PR #54, opened 2017-02-23
Add examples of setting group triggers and members

zgalor:
Examples that set autoresolving triggers for ops metrics:
       docker.ping < 1
       kernel.all.cpu.idle < 5
       openshift.node.process.count != 1

PR #58, opened 2017-03-16
Group example playbooks in "examples" dir

dkorn:
Notice it is now required to add `-M library/` when running the example playbooks


dkorn:
@cben @simon3z @zgalor PTAL

examples/example_add_openshift_provider_playbook.retry
```diff
@@ -0,0 +1 @@
+localhost
```
cben: did you git add this file by mistake?

cben:
LGTM

examples/example_add_openshift_provider_playbook.retry
dkorn: I did, removed now.
nice catch :)

zgalor:
LGTM :+1: 
Now you can merge my PR https://github.com/dkorn/manageiq-ansible-module/pull/54 :)

dkorn:
@zgalor I don't think I'll merge #54 for two reasons:
a. it should be part of #51, as it refers to the CM-OPS deployment
b. it should be submitted as an example to https://github.com/dkorn/hawkular-alerts-module

PR #59, opened 2017-03-17
Add a LICENSE.txt

dkorn:


dkorn:
@simon3z I added the MIT License, is there any reason I should use the Apache License 2.0?

simon3z:
> @simon3z I added the MIT License, is there any reason I should use the Apache License 2.0?

@dkorn I think we'll want (part of) this to end in the ansible core repository (which IIRC is using GPLv3) so we should use the same license or a compatible one.
WRT naming of the file I remember that `COPYING` (instead of `LICENSE.txt`) was widely used.

PR #62, opened 2017-03-22
Fix the documentation inside the modules

### LINKED ISSUE #56: Fix the documentation on the modules
`ansible-doc` command should show documentation on Ansible modules.
It fails on all modules with a useless error message.

We should at least:

- [ ] check the yaml validity

- [ ] add top level `description` (in addition to the existing `short_descriptions`) 

cc @cben

dkorn:
Running `ansible-doc --module-path=library/ <module_name>.py` should output the module docs.
This change fixes few small issues preventing the docs display.

fixes: #56


dkorn:
@cben @enoodle @ivanhorvath @simon3z PTAL

cben (approved):

PR #63, opened 2017-03-22
Provider custom CA support backward compatibility

### LINKED ISSUE #61: Adding provider to Euwe fails even if provider_verify_ssl is set to false
Attempting to run add_provider.yml with an Openshift provider and getting the following error:

"msg": "Failed to add provider. Error: APIException(\"Api::BadRequestError: Could not create the new provider - unknown attribute 'certificate_authority' for Endpoint.\",)"

The yml looks like this:

--- 
- 
  hosts: localhost
  tasks: 
    - 
      manageiq_provider: 
        hawkular_hostname: <hostname>
        hawkular_port: 443
        metrics: "True"
        miq_password: 
        miq_url: <url>
        miq_username: 
        miq_verify_ssl: "False"
        name: "CI OSE"
        provider_api_auth_token: <token>
        provider_api_hostname: <hostname>
        provider_api_port: 8443
        provider_type: openshift-origin
        state: present
      name: "Add Openshift Containers Provider to ManageIQ"
      register: result
    - 
      debug: var=result

More info here:
http://pastebin.test.redhat.com/466033

dkorn:
Following a recent change in manageiq PR #55 added support for provider TLS and custom CA.
Adding a provider with an auth_key_config endpoint was failing on previous manageiq versions, due to unknown attribute 'certificate_authority'.

This change handles the issue, fixes: #61

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ The `manageiq_provider` module currently supports adding, updating and deleting 
 Example playbooks [add_openshift_provider.yml](add_openshift_provider.yml), [add_amazon_provider.yml](add_amazon_provider.yml) and [add_hawkular_datawarehouse_provider.yml](add_hawkular_datawarehouse_provider.yml) are provided.
 To update an existing provider pass the changed values together with the required parameters. To delete a provider change `state=absent`.  
 SSL verification for HTTPS requests between ManageIQ and the provider is enabled by default. To ignore pass `provider_verify_ssl: false`.
-To use a self-signed certificate pass: `provider_ca_path: '/path/to/certfile'`.  
+To use a self-signed certificate pass: `provider_ca_path: '/path/to/certfile'`. To remove a previously defined ca pass `""` (empty string). In case the parameter is passed with null or omitted the `certificate_authority` field will be left unmodified (unset on creation). `provider_ca_path` must be omitted with ManageIQ Euwe / CFME 5.7 or earlier releases.
 After addition or update, each endpoint authentication is validated, a process which can take up to 50 seconds before timeout.
 If all authentications are valid the provider's inventory is refreshed.
 
--- a/library/manageiq_provider.py
+++ b/library/manageiq_provider.py
@@ -85,6 +85,11 @@ options:
   provider_ca_path:
     description:
       - the path to the ca file
+      - to remove a previously defined ca pass "" (empty string)
+      - in case the parameter is passed with null or omitted the
+        certificate_authority field will be left unmodified
+        (unset on creation).
+      - must be omitted with ManageIQ Euwe / CFME 5.7 or earlier releases
     required: false
     default: null
   metrics:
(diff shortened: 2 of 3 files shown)


dkorn:
@cben @enoodle @ivanhorvath @simon3z please review

cc @pavelzag

README.md
```diff
@@ -32,7 +32,7 @@ The `manageiq_provider` module currently supports adding, updating and deleting
 Example playbooks [add_openshift_provider.yml](add_openshift_provider.yml), [add_amazon_provider.yml](add_amazon_provider.yml) and [add_hawkular_datawarehouse_provider.yml](add_hawkular_datawarehouse_provider.yml) are provided.
 To update an existing provider pass the changed values together with the required parameters. To delete a provider change `state=absent`.  
 SSL verification for HTTPS requests between ManageIQ and the provider is enabled by default. To ignore pass `provider_verify_ssl: false`.
-To use a self-signed certificate pass: `provider_ca_path: '/path/to/certfile'`.  
+To use a self-signed certificate pass: `provider_ca_path: '/path/to/certfile'`. To remove a previously defined ca pass `""` (empty string). In case the parameter is passed with null or omitted the endpoint will be generated without a `certificate_authority`.
```
cben: This describes what "generate_endpoint" function does, but is confusing to user.
To user, "the `certificate_authority` field will be left unmodified (unset on creation)".

Perhaps also mention it "must be omitted with ManageIQ Euwe / CFME 5.7 or earlier releases, causes an error".

library/manageiq_provider.py
```diff
@@ -85,6 +85,9 @@
   provider_ca_path:
     description:
       - the path to the ca file
+      - to remove a previously defined ca pass "" (empty string)
+      - in case the parameter is passed with null or omitted the
+        endpoint will be generated without a certificate_authority
```
cben: Same comment as in README

library/manageiq_provider.py
```diff
@@ -356,12 +359,19 @@ def generate_auth_key_config(self, role, authtype, hostname, port, token, provid
         if provider_ca_path:
             with open(provider_ca_path, 'r') as provider_ca_file:
                 provider_ca_content = provider_ca_file.read()
+        elif provider_ca_path == "":
+            provider_ca_content = ""
```
cben: I don't like this None / something else structure, with if is not None below.
Can you put `endpoint = {...` first, then do this if/elif setting `endpoint['endpoint']['certificate_authority']`?

Also don't call the var `endpoint` if it has 'endpoint' and 'authentication' inside.  How about `config`?

library/manageiq_provider.py
dkorn: done, thanks!

README.md
dkorn: Done, thanks!

library/manageiq_provider.py
dkorn: great suggestion! making the change, thanks :)

cben (approved):

PR #66, opened 2017-03-28
add initiate_refresh and validate_provider_auth flags to openshift provider module

enoodle:
This will allow adding providers that are currently offline but I know that they are valid (VMs are currently off, network problems etc)

### DIFF
--- a/library/manageiq_provider.py
+++ b/library/manageiq_provider.py
@@ -105,6 +105,16 @@ options:
       - the port used for hawkular metrics
     required: false
     default: null
+  validate_provider_auth:
+    description:
+      - disable the provider authentication validation
+    required: false
+    default: true
+  initiate_refresh:
+    description:
+      - disable the provider inventory refresh initiation
+    required: false
+    default: true
 '''
 
 EXAMPLES = '''
@@ -427,7 +437,8 @@ class ManageIQProvider(object):
                     if field in endpoint and endpoint[field] is None:
                         del endpoint[field]
 
-    def add_or_update_provider(self, provider_name, provider_type, endpoints, zone, provider_region):
+    def add_or_update_provider(self, provider_name, provider_type, endpoints, zone, provider_region,
+            validate_provider_auth = True, initiate_refresh = True):
         """ Adds a provider to manageiq or update its attributes in case
         a provider with the same name already exists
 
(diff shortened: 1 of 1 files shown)


enoodle:
@dkorn PTAL

library/manageiq_provider.py
```diff
@@ -109,6 +109,11 @@
       - the port used for hawkular metrics
     required: false
     default: null
+  skip_refresh_and_validation:
```
dkorn: regarding the flag name:
a. validation comes chronologically prior to refresh, so we should probably switch the order unless
b. we split it into 2 flags, so that the refresh depends on the validation (can't happen without it) and can also be skipped
c. if we decide to go with one flag, find a shorter name

@cben please also advise

library/manageiq_provider.py
```diff
@@ -109,6 +109,11 @@
...
+    description:
+      - disable refreshing and validating of the provider
```
dkorn: `disables the provider authentication validation and its inventory refresh`

library/manageiq_provider.py
```diff
@@ -432,16 +437,20 @@ def add_or_update_provider(self, provider_name, provider_type, endpoints, zone,
...
-                authtypes_to_verify.append(e['authentication']['authtype'])
-        result, details = self.verify_authenticaion_validation(provider_id, old_validation_details, authtypes_to_verify)
+        if not skip_refresh_and_validation:
+            authtypes_to_verify = []
+            for e in endpoints:
+                if e['endpoint']['role'] in roles_with_changes:
+                    authtypes_to_verify.append(e['authentication']['authtype'])
+            result, details = self.verify_authenticaion_validation(provider_id, old_validation_details, authtypes_to_verify)
+        else:
+            result = "Valid"
```
dkorn: I don't think `"Valid"` is the right term to describe the result of the authentication in case it was skipped.
Consider adding another option of `"Skipped"` and:

``` python 
elif result == "Skipped":
```

library/manageiq_provider.py
```diff
@@ -478,7 +487,8 @@ def main():
             secret_access_key=dict(required=False, type='str', no_log=True),
             metrics=dict(required=False, type='bool', default=False),
             hawkular_hostname=dict(required=False),
-            hawkular_port=dict(required=False)
+            hawkular_port=dict(required=False),
+            skip_refresh_and_validation=dict(required=False, default=False)
```
dkorn: please add 
`type='bool'`

dkorn (commented):
Thanks for the change!
Few comments :)

cben:
Haven’t looked at code yet.

Interface idea: what do you think of `state: present` vs `state: validated` vs `state: refreshed` as the interface?
​

enoodle:
@dkorn what do you think about Beni's proposal?

dkorn:
@enoodle @cben I prefer not to go with the interface idea @cben suggested for several reasons:

a. it isn't really the "Ansible way", meaning working with the present-absent states.
b. I'm trying to maintain consistency with the other manageiq modules in this repo
c. currently, present state holds adding, validating and refreshing - the most common use case. adding the other options confuses the user in terms of state hierarchy.
d. you can't really have a refresh**ed** state, more like refresh**ing**, since we are just triggering a refresh task as it may take a long time. Having said that, the same goes for the deleting task in absent.

I think separating into two flags will be straight-forward to use and easy to implement

enoodle:
@dkorn :+1: Ill work on that

enoodle:
@dkorn  I have split into two flags with positive meanings and `True` as default.

library/manageiq_provider.py
```diff
@@ -109,6 +109,16 @@
...
+  validate_provider_auth:
+    description:
+      - disable the provider authentication validation
+    required: false
+    default: false
+  initiate_refresh:
+    description:
+      - disable the provider inventory refresh initiation
+    required: false
+    default: false
```
cben: ^^ These 2 actual defaults are True

cben (commented):
LGTM aside from one doc catch.

I'd love to split `add_or_update_provider` up but don't have a concrete suggestion at this moment, will send a PR if I figure a way.

library/manageiq_provider.py
enoodle: fixed

cben:
LGTM :+1:.  Please rename commit & PR to reflect the split flag names.

And there is now a conflict, you'll have to rebase.

enoodle:
@cben done. I changed the PR title and commit message.

PR #67, opened 2017-03-29
Send security_protocol field with provider endpoint

zgalor:
Deduce security_protocol from provider_ssl parameters, and send it to manageiq.
Also check if provider_verify_ssl and provider_ca_content differ in playbook from provider and update accordingly.

### DIFF
--- a/library/manageiq_provider.py
+++ b/library/manageiq_provider.py
@@ -85,12 +85,8 @@ options:
     choices: ['True', 'False']
   provider_ca_path:
     description:
-      - the path to the ca file
-      - to remove a previously defined ca pass "" (empty string)
-      - in case the parameter is passed with null or omitted the
-        certificate_authority field will be left unmodified
-        (unset on creation).
-      - must be omitted with ManageIQ Euwe / CFME 5.7 or earlier releases
+      - path to a file with certificate authoritie(s) to trust, in PEM format
+      - to remove a previously defined ca pass null or omit
     required: false
     default: null
   metrics:
@@ -255,7 +251,15 @@ class ManageIQ(object):
 
         return "Timed out", details
 
-    def required_updates(self, provider_id, endpoints, zone_id, provider_region):
+    def get_provider_config(self, provider_id):
+        """ get the endpoint content of existing provider from manageiq API"""
+        try:
+            result = self.client.get('{providers_url}/{id}/?attributes=endpoints'.format(providers_url=self.providers_url, id=provider_id))
+            return result
+        except Exception as e:
+            self.module.fail_json(msg="Failed to get provider data. Error: {!r}".format(e))
+
+    def required_updates(self, provider_id, endpoints, zone_id, provider_region, existing_config):
         """ Checks whether an update is required for the provider
 
         Returns:
@@ -268,11 +272,6 @@ class ManageIQ(object):
                                 that will contain all the changed endpoints
                                 and their values.
         """
-        try:
-            result = self.client.get('{providers_url}/{id}/?attributes=endpoints'.format(providers_url=self.providers_url, id=provider_id))
-        except Exception as e:
-            self.module.fail_json(msg="Failed to get provider data. Error: {!r}".format(e))
-
         def host_port_ssl(endpoint):
             return {'hostname': endpoint.get('hostname'),
                     'port': endpoint.get('port'),
@@ -281,19 +280,19 @@ class ManageIQ(object):
                     'security_protocol': endpoint.get('security_protocol')}
 
         desired_by_role = {e['endpoint']['role']: host_port_ssl(e['endpoint']) for e in endpoints}
-        result_by_role = {e['role']: host_port_ssl(e) for e in result['endpoints']}
-        existing_provider_region = result.get('provider_region') or None
-        if result_by_role == desired_by_role and result['zone_id'] == zone_id and existing_provider_region == provider_region:
+        existing_by_role = {e['role']: host_port_ssl(e) for e in existing_config['endpoints']}
+        existing_provider_region = existing_config.get('provider_region') or None
+        if existing_by_role == desired_by_role and existing_config['zone_id'] == zone_id and existing_provider_region == provider_region:
             return {}
         updated = {role: {k: v for k, v in ep.items()
-                          if k not in result_by_role[role] or v != result_by_role[role][k]}
+                          if k not in existing_by_role[role] or v != existing_by_role[role][k]}
                    for role, ep in desired_by_role.items()
-                   if role in result_by_role and ep != result_by_role[role]}
+                   if role in existing_by_role and ep != existing_by_role[role]}
         added = {role: ep for role, ep in desired_by_role.items()
-                 if role not in result_by_role}
-        removed = {role: ep for role, ep in result_by_role.items()
+                 if role not in existing_by_role}
+        removed = {role: ep for role, ep in existing_by_role.items()
                    if role not in desired_by_role}
-        if result['zone_id'] != zone_id:
+        if existing_config['zone_id'] != zone_id:
             updated['zone_id'] = zone_id
         if existing_provider_region != provider_region:
             updated['provider_region'] = provider_region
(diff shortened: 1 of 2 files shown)


zgalor:
@cben @dkorn PTAL

zgalor:
wait, I'll fix the tests...

library/manageiq_provider.py
```diff
@@ -274,9 +274,12 @@ def required_updates(self, provider_id, endpoints, zone_id, provider_region):
             self.module.fail_json(msg="Failed to get provider data. Error: {!r}".format(e))
 
         def host_port(endpoint):
```
dkorn: please rename this method as it now takes care of other attributes

library/manageiq_provider.py
```diff
@@ -370,6 +373,14 @@ def generate_auth_key_config(self, role, authtype, hostname, port, token, provid
         elif provider_ca_path == "":
             config['endpoint']['certificate_authority'] = ""
 
+        if provider_verify_ssl:
+            if config['endpoint'].get('certificate_authority', None):
```
dkorn: nit: you can remove the `None` as it's the default value

dkorn (changes requested):
Thanks for the PR!

This brings the questions of how to handle create/update in cases where the CA isn't needed. 
IMO, sending the attribute with None (null) isn't the correct approach here, I prefer adding the field only when needed. The only exception is when we need to remove the CA in an update and then sending the attribute with None is mundatory.

zgalor:
@dkorn I made the changes you requested. 

Regarding backward compatibility: If we decide not to support update of certificate_authority (specifically not allowing to remove it) - we can simply not send it if it's null or empty - which will not trigger API errors in older manageiq versions.
Update of that field will only be allowed through removal of the provider and recreation. I must note that anyway manageiq doesn't update that field when sent empty, so I'm not sure the module should support it either.

@cben @dkorn what do you think?

dkorn:
> Regarding backward compatibility: If we decide not to support update of certificate_authority (specifically not allowing to remove it) - we can simply not send it if it's null or empty - which will not trigger API errors in older manageiq versions.
> Update of that field will only be allowed through removal of the provider and recreation. I must note that anyway manageiq doesn't update that field when sent empty, so I'm not sure the module should support it either.
> 
> @cben @dkorn what do you think?

As we discussed, we want to support `certificate_authority` update.
The only _tricky_ part is to updating when the certificate_authority is not needed. In this case we will check if the field certificate_authority exists in the endpoint data and act accordingly.

dkorn (approved):

dkorn:
Approved only not to prevent merging in case @cben approves

zgalor:
I added another commit that adds support for CFME 5.7 / ManageIQ Euwe or older versions, where endpoint API doesn't have the certificate_authority field.
The way it works:
On endpoint creation, don't send certificate_authority field if null or omitted.
On update, send a null certificate_authority field only if the current provider already has a non-null-value certificate_authority field.

I added unit tests for the filtering method and fixed existing unit tests to align with new behavior.
This has also been tested against upstream ManageIQ and CFME 5.7.

@cben can you please review/approve?

library/manageiq_provider.py
```diff
@@ -86,11 +86,8 @@
   provider_ca_path:
     description:
       - the path to the ca file
```
cben: can you improve this line?  "path to a file with certificate authoritie(s) to trust, in PEM format"

library/manageiq_provider.py
```diff
@@ -86,11 +86,8 @@
...
-      - to remove a previously defined ca pass "" (empty string)
+      - to remove a previously defined ca pass null or omit
       - in case the parameter is passed with null or omitted the
```
cben: drop this line

library/manageiq_provider.py
```diff
@@ -268,19 +273,17 @@ def required_updates(self, provider_id, endpoints, zone_id, provider_region):
...
-        if result_by_role == desired_by_role and result['zone_id'] == zone_id and existing_provider_region == provider_region:
+        def host_port_ssl(endpoint):
+            return {'hostname': endpoint.get('hostname'),
+                    'port': endpoint.get('port'),
+                    'verify_ssl': endpoint.get('verify_ssl'),
+                    'certificate_authority': endpoint.get('certificate_authority'),
+                    'security_protocol': endpoint.get('security_protocol')}
+
+        desired_by_role = {e['endpoint']['role']: host_port_ssl(e['endpoint']) for e in endpoints}
+        result_by_role = {e['role']: host_port_ssl(e) for e in existing_config['endpoints']}
```
cben: Consider s/desired_by_role/existing_by_role/.  "result" is unclear after this refactoring.

library/manageiq_provider.py
```diff
@@ -367,8 +370,17 @@ def generate_auth_key_config(self, role, authtype, hostname, port, token, provid
...
                 provider_ca_content = provider_ca_file.read()
                 config['endpoint']['certificate_authority'] = provider_ca_content
-        elif provider_ca_path == "":
-            config['endpoint']['certificate_authority'] = ""
+        else:
+            config['endpoint']['certificate_authority'] = None
+
+        # deduce security_protocol from provider_verify_ssl and certificate_authority
+        if provider_verify_ssl:
+            if config['endpoint'].get('certificate_authority'):
```
cben: I'd maybe change this to `if provider_ca_path:`.  Looking at config was to rely on logic above when it was more complicated.

library/manageiq_provider.py
```diff
@@ -402,6 +414,20 @@ def delete_provider(self, provider_name):
         else:
             return dict(task_id=None, changed=self.changed, msg="Provider {provider_name} doesn't exist".format(provider_name=provider_name))
 
+    def filter_empty_unrequired_fields(self, configs, existing_endpoints, fields):
+        """
+        Only update fields that already exist in the endpoint with empty values.
+        :param endpoints: New endpoints configutarion
```
cben: out of date, it's config.
also please document it mutates it in-place.

library/manageiq_provider.py
```diff
@@ -402,6 +414,20 @@ def delete_provider(self, provider_name):
         else:
             return dict(task_id=None, changed=self.changed, msg="Provider {provider_name} doesn't exist".format(provider_name=provider_name))
 
+    def filter_empty_unrequired_fields(self, configs, existing_endpoints, fields):
```
cben: Not sure it's worth being generic on `fields`.

This mutates `configs` argument, which is rare for a python function.  Perhaps calling it "delete_..." would make it clearer.

library/manageiq_provider.py
```diff
@@ -415,7 +441,14 @@ def add_or_update_provider(self, provider_name, provider_type, endpoints, zone,
         # check if provider with the same name already exists
         provider_id = self.find_provider_by_name(provider_name)
         if provider_id:  # provider exists
-            updates = self.required_updates(provider_id, endpoints, zone_id, provider_region)
+            existing_config = self.get_provider_config(provider_id)
+
+            # ManageIQ Euwe / CFME 5.7 API and older versions don't support certificate authority field in endpoint.
+            # If it didn't exist in existing provider endpoints, we can remove null/empty certificate_authority from endpoints we want to update.
```
cben: there is a subtle distinction between "didn't exist in provider" vs "wasn't returned from API".
Worth explaining that wasn't returned means either unsupported or supported but currently null.

tests/test_manageiq_provider.py
```diff
@@ -252,8 +261,28 @@ def test_generate_auth_key_config(miq):
...
+                             'field_that_should_be_sent': 'new_data',
+                             'empty_field_that_should_be_sent': None}}]
+
+    existing_endpoints = [{'field_that_isnt_updated': 'value',
+                           'field_that_should_be_sent': 'old_data',
+                           'empty_field_that_should_be_sent': 'some_data'}]
+
+    miq.filter_empty_unrequired_fields(configs, existing_endpoints, fields)
+    assert configs == [{'endpoint': {'field_that_should_be_sent': 'new_data',
+                                     'empty_field_that_should_be_sent': None}}]
```
cben: On one hand, awesome test, puts `fields` to neat use :heart: 
OTOH, it covers what the function *does* but it's hard to see *why* it achieves the desired backward-compat.
Perhaps scenario-like names like
```
unsupported_field_leaving_empty
unsupported_field_trying_to_set
supported_empty_leaving_empty
supported_empty_setting
supported_non_empty_clearing
supported_non_empty_setting
```
could help?
(The 3x2 matrix is bigger because some scenarios collapse to same function behavior.)

cben (commented):
Thanks for tackling this mess!
Overall good, small suggestions, feel free to disagree.

library/manageiq_provider.py
zgalor: I wouldn't use `delete` on this one because it implies we are deleting some fields from the provider, which is the exact opposite of the intention of this method.
I will rename to `filter_unsupported_fields_from_config` which is clearer.

zgalor:
@cben I applied all the changes you requested (all but one rename suggestion, I replied to the comment).
Please approve.

cben (approved):

cben:
Using my merge powers without @dkorn as he's traveling and this matches what we discussed together.

PR #68, opened 2017-04-03
Alert Definition module

dkorn:
Example playbook can be found in `examples/create_alert.yml`

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ To unassign a tag from a resource change `state=absent`.
 The `manageiq_alert` module supports adding, updating and deleting alerts in manageiq.  
 Example playbook [create_alert.yml](examples/create_alert.yml) is provided.  
 To update an existing alert pass the changed values together with the required parameters. To delete an alert change `state=absent`.
-The following ManageIQ entities supports alerts: node, vm, host, storage, cluster, ems, miq_server and middleware_server.
+The following ManageIQ entities supports alerts: container_node, vm, host, storage, cluster, ems, miq_server and middleware_server.
 
 
 
--- a/README.md
+++ b/README.md
@@ -69,6 +69,12 @@ Each assigned tag must be a part of a unique category.
 An example playbook [assign_tag.yml](assign_tag.yml) is provided.  
 To unassign a tag from a resource change `state=absent`.
 
+### manageiq_alert module
+
+The `manageiq_alert` module supports adding, updating and deleting alerts in manageiq.  
+Example playbook [create_alert.yml](examples/create_alert.yml) is provided.  
+To update an existing alert pass the changed values together with the required parameters. To delete an alert change `state=absent`.
+
 
 
 ## Using Environment Variables
--- a/README.md
+++ b/README.md
@@ -74,6 +74,7 @@ To unassign a tag from a resource change `state=absent`.
 The `manageiq_alert` module supports adding, updating and deleting alerts in manageiq.  
 Example playbook [create_alert.yml](examples/create_alert.yml) is provided.  
 To update an existing alert pass the changed values together with the required parameters. To delete an alert change `state=absent`.
+The following ManageIQ entities supports alerts: node, vm, host, storage, cluster, ems, miq_server and middleware_server.
 
 
 
--- a/README.md
+++ b/README.md
@@ -74,6 +74,7 @@ To unassign a tag from a resource change `state=absent`.
 The `manageiq_alert` module supports adding, updating and deleting alerts in manageiq.  
 Example playbook [create_alert.yml](examples/create_alert.yml) is provided.  
 To update an existing alert pass the changed values together with the required parameters. To delete an alert change `state=absent`.
+Alert expression in ManageIQ can have one of two types: `miq_expression` (`MiqExpression`) or `hash`. By default the expression type is `miq_expression`, to change the type use the `expression_type` field.
 The following ManageIQ entities supports alerts: container_node, vm, host, storage, cluster, ems, miq_server and middleware_server.
 
 
--- a/examples/create_alert.yml
+++ b/examples/create_alert.yml
@@ -1,0 +1,21 @@
+---
+- hosts: localhost
+
+  tasks:
+  - name: Create an alert in ManageIQ
+    manageiq_alert:
+      miq_url: http://miq.example.com
+      miq_username: admin
+      miq_password: secret
+      miq_verify_ssl: false
+      description: Test Alert 01
+      options:
+        notifications:
+          delay_next_evaluation: 600
+          evm_event: {}
+      db: ContainerNode
+      expression:
+        eval_method: dwh_generic
+        mode: internal
+      enabled: true
+      state: present
--- a/examples/create_alert.yml
+++ b/examples/create_alert.yml
@@ -13,9 +13,11 @@
         notifications:
           delay_next_evaluation: 0
           evm_event: {}
-      entity: node
+      entity: container_node
       expression:
         eval_method: dwh_generic
         mode: internal
+        options: {}
+      expression_type: hash
       enabled: true
       state: present
--- a/examples/create_alert.yml
+++ b/examples/create_alert.yml
@@ -13,7 +13,7 @@
         notifications:
           delay_next_evaluation: 600
           evm_event: {}
-      db: ContainerNode
+      entity: node
       expression:
         eval_method: dwh_generic
         mode: internal
--- a/examples/create_alert.yml
+++ b/examples/create_alert.yml
@@ -11,7 +11,7 @@
       description: Test Alert 01
       options:
         notifications:
-          delay_next_evaluation: 600
+          delay_next_evaluation: 0
           evm_event: {}
       entity: node
       expression:
--- a/library/manageiq_alert.py
+++ b/library/manageiq_alert.py
@@ -49,6 +49,12 @@ options:
       - the expression to evaluate
     required: false
     default: null
+  expression_type:
+    description:
+      - the type of the alert expression
+    required: false
+    choices: ['miq_expression', 'hash']
+    default: miq_expression
   options:
     description:
       - Additional alert options, including the notification type and frequency
@@ -82,6 +88,7 @@ EXAMPLES = '''
     expression:
       eval_method: dwh_generic
       mode: internal
+    expression_type: hash
     enabled: true
     state: present
     miq_url: 'http://localhost:3000'
@@ -150,7 +157,7 @@ class ManageIQAlert(object):
         self.changed = True
         return dict(changed=self.changed, msg=result['message'])
 
-    def alert_update_required(self, alert_id, description, expression, miq_entity, options, enabled):
+    def alert_update_required(self, alert_id, description, expression, expression_type, miq_entity, options, enabled):
         """ Returns true if the expression, miq_entity, options, or enabled passed for
             the alert differ from the alert's existing ones, False otherwise.
         """
@@ -160,8 +167,12 @@ class ManageIQAlert(object):
         except Exception as e:
             self.module.fail_json(msg="Failed to get alert {description} details. Error: {error}".format(description=description, error=e))
 
-        # remove None values from expression and options dicts
-        current_expression = {k: v for k, v in result['expression']['exp'].items() if v is not None}
+        # remove None values from expression and options dicts, if needed
+        # TODO (dkorn): use the expression_type from the response, once supported
+        if expression_type == 'miq_expression':
+            current_expression = {k: v for k, v in result['expression']['exp'].items() if v is not None}
+        else:
+            current_expression = result['expression']
         current_options = {k: v for k, v in result['options'].items() if v is not None}
 
         attributes_tuples = [(current_expression, expression), (result['db'], miq_entity), (current_options, options), (result['enabled'], enabled)]
@@ -170,20 +181,21 @@ class ManageIQAlert(object):
                 return True
         return False
 
-    def update_alert_if_required(self, alert_id, description, expression, miq_entity, options, enabled):
+    def update_alert_if_required(self, alert_id, description, expression, expression_type, miq_entity, options, enabled):
         """Updates the alert in manageiq.
 
         Returns:
             Whether or not a change took place and a message describing the
             operation executed.
         """
-        if not self.alert_update_required(alert_id, description, expression, miq_entity, options, enabled):
+        if not self.alert_update_required(alert_id, description, expression, expression_type, miq_entity, options, enabled):
             return dict(
                 changed=self.changed,
                 msg="Alert {description} already exist, no need for updates".format(description=description))
 
         url = '{api_url}/alert_definitions/{alert_id}'.format(api_url=self.api_url, alert_id=alert_id)
-        resource = {'description': description, 'expression': expression, 'db': miq_entity,
+        resource = {'description': description, 'expression': expression,
+                    'expression_type': expression_type, 'db': miq_entity,
                     'options': options, 'enabled': enabled}
         try:
             result = self.client.post(url, action='edit', resource=resource)
@@ -194,7 +206,7 @@ class ManageIQAlert(object):
             changed=self.changed,
             msg="Successfully updated alert {description}: {alert_details}".format(description=description, alert_details=result))
 
-    def create_alert(self, description, expression, miq_entity, options, enabled):
+    def create_alert(self, description, expression, expression_type, miq_entity, options, enabled):
         """Creates the alert in manageiq.
 
         Returns:
@@ -203,7 +215,8 @@ class ManageIQAlert(object):
         """
         url = '{api_url}/alert_definitions/'.format(api_url=self.api_url)
         resource = {'description': description, 'expression': expression,
-                    'db': miq_entity, 'options': options, 'enabled': enabled}
+                    'expression_type': expression_type, 'db': miq_entity,
+                    'options': options, 'enabled': enabled}
         try:
             result = self.client.post(url, action='create', resource=resource)
             self.changed = True
@@ -213,7 +226,7 @@ class ManageIQAlert(object):
         except Exception as e:
             self.module.fail_json(msg="Failed to create alert {description}: {error}".format(description=description, error=e))
 
-    def create_or_update_alert(self, description, expression, entity, options, enabled):
+    def create_or_update_alert(self, description, expression, expression_type, entity, options, enabled):
         """ Create or update an alert in manageiq.
 
         Returns:
@@ -223,9 +236,9 @@ class ManageIQAlert(object):
         miq_entity = ManageIQAlert.supported_entities[entity]
         alert_id = self.find_alert_by_description(description)
         if alert_id:  # alert already exist
-            return self.update_alert_if_required(alert_id, description, expression, miq_entity, options, enabled)
+            return self.update_alert_if_required(alert_id, description, expression, expression_type, miq_entity, options, enabled)
         else:
-            return self.create_alert(description, expression, miq_entity, options, enabled)
+            return self.create_alert(description, expression, expression_type, miq_entity, options, enabled)
 
 
 def main():
@@ -238,6 +251,9 @@ def main():
                                  'middleware_server']),
             options=dict(required=False, type='dict'),
             expression=dict(required=False, type='dict'),
+            expression_type=dict(required=False, type='str',
+                                 default='miq_expression',
+                                 choices=['miq_expression', 'hash']),
             enabled=dict(require=False, type='bool', default=True),
             state=dict(required=True, type='str',
                        choices=['present', 'absent']),
@@ -256,22 +272,24 @@ def main():
         if module.params[arg] in (None, ''):
             module.fail_json(msg="missing required argument: {}".format(arg))
 
-    miq_url        = module.params['miq_url']
-    miq_username   = module.params['miq_username']
-    miq_password   = module.params['miq_password']
-    miq_verify_ssl = module.params['miq_verify_ssl']
-    ca_bundle_path = module.params['ca_bundle_path']
-    description    = module.params['description']
-    entity         = module.params['entity']
-    options        = module.params['options']
-    expression     = module.params['expression']
-    enabled        = module.params['enabled']
-    state          = module.params['state']
+    miq_url         = module.params['miq_url']
+    miq_username    = module.params['miq_username']
+    miq_password    = module.params['miq_password']
+    miq_verify_ssl  = module.params['miq_verify_ssl']
+    ca_bundle_path  = module.params['ca_bundle_path']
+    description     = module.params['description']
+    entity          = module.params['entity']
+    options         = module.params['options']
+    expression      = module.params['expression']
+    expression_type = module.params['expression_type']
+    enabled         = module.params['enabled']
+    state           = module.params['state']
 
     manageiq = ManageIQAlert(module, miq_url, miq_username, miq_password, miq_verify_ssl, ca_bundle_path)
     if state == "present":
         res_args = manageiq.create_or_update_alert(description, expression,
-                                                   entity, options, enabled)
+                                                   expression_type, entity,
+                                                   options, enabled,)
     if state == "absent":
         res_args = manageiq.delete_alert(description)
 
--- a/library/manageiq_alert.py
+++ b/library/manageiq_alert.py
@@ -21,15 +21,27 @@ options:
     description:
       - manageiq password
     default: MIQ_PASSWORD env var if set. otherwise, it is required to pass it
+  miq_verify_ssl:
+    description:
+      - whether SSL certificates should be verified for HTTPS requests
+    required: false
+    default: True
+    choices: ['True', 'False']
+  ca_bundle_path:
+    description:
+      - the path to a CA_BUNDLE file or directory with certificates
+    required: false
+    default: null
   description:
     description:
       - the alert definition description in manageiq
     required: false
     default: null
-  db:
+  entity:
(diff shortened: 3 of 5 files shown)


dkorn:
@cben @enoodle @ivanhorvath @zgalor @simon3z @moolitayer please review

zgalor:
I don't know if hawkular API allows editing this, but it would be cool if we could set alert_status "open/resolved" with the playbook

zgalor:
I am not sure about the usage, but I believe we will need the option to add tags on the alert for manageiq to show them. @moolitayer keep me honest here

library/manageiq_alert.py
```diff
@@ -0,0 +1,272 @@
...
+  state:
+    description:
+      - the state of the alert definition
+      - On present, it will create the alert if it does not exist, or update it
+        if the associated data is different
+      - On absent, it will delete the alert if it exists
+    required: false
+    choices: ['present', 'absent']
+    default: 'present'
+  miq_verify_ssl:
```
cben: nitpick: move connection-related options together. ca_bundle_path too.

library/manageiq_alert.py
```diff
@@ -0,0 +1,272 @@
...
+      - manageiq password
+    default: MIQ_PASSWORD env var if set. otherwise, it is required to pass it
+  description:
+    description:
+      - the alert definition description in manageiq
+    required: false
+    default: null
+  db:
+    description:
+      - the db entry of the entity to base the alert on in manageiq
```
cben: I don't quite like the name `db`, nor the fact it takes internal class names as values.
How about something like `resource` or `entity`?  And perhaps a mapping hash like you did in a couple places.

library/manageiq_alert.py
dkorn: Done, thanks!

library/manageiq_alert.py
dkorn: Done, thanks!

library/manageiq_alert.py
```diff
@@ -0,0 +1,282 @@
...
+    user           - the username in manageiq
+    password       - the user password in manageiq
+    miq_verify_ssl - whether SSL certificates should be verified for HTTPS requests
+    ca_bundle_path - the path to a CA_BUNDLE file or directory with certificates
+    """
+
+    supported_entities = {
+        'node': 'ContainerNode', 'vm': 'Vm', 'miq_server': 'MiqServer', 'host': 'Host',
+        'storage': 'Storage', 'cluster': 'EmsCluster', 'ems': 'ExtManagementSystem',
+        'miq_server': 'MiqServer', 'middleware_server': 'MiddlewareServer'
```
zgalor: I would rename this to `entity_to_miq_entity_map` and change all the leftover `db` vars and parameters to `miq_entity`

library/manageiq_alert.py
```diff
@@ -0,0 +1,282 @@
...
+            self.module.fail_json(msg="Failed to create alert {description}: {error}".format(description=description, error=e))
+
+    def create_or_update_alert(self, description, expression, entity, options, enabled):
+        """ Create or update an alert in manageiq.
+
+        Returns:
+            Whether or not a change took place and a message describing the
+            operation executed.
+        """
+        db = ManageIQAlert.supported_entities[entity]
```
zgalor: I would rename `db` to `miq_entity`

tests/test_manageiq_alert.py
```diff
@@ -0,0 +1,214 @@
...
+
+MANAGEIQ_HOSTNAME = "http://miq.example.com"
+DESCRIPTION = "Test Alert 01"
+ALERT_ID = "17"
+EXPRESSION = {
+    "eval_method": "dwh_generic",
+    "mode": "internal",
+    "options": {}
+}
+DB = "ContainerNode"
```
zgalor: rename `DB` to `ENTITY`

zgalor (commented):
For consistency with the last change, I would also change other `db` internal usages to `entity` or `miq_entity` by context

library/manageiq_alert.py
dkorn: I prefer to maintain consistency with other modules using the naming `supported_entities`

tests/test_manageiq_alert.py
dkorn: Done, thanks!

library/manageiq_alert.py
dkorn: will do, thanks!

library/manageiq_alert.py
zgalor: :+1:

library/manageiq_alert.py
```diff
@@ -0,0 +1,282 @@
...
+            Whether or not a change took place and a message describing the
+            operation executed.
+        """
+        if not self.alert_update_required(alert_id, description, expression, miq_entity, options, enabled):
+            return dict(
+                changed=self.changed,
+                msg="Alert {description} already exist, no need for updates".format(description=description))
+
+        url = '{api_url}/alert_definitions/{alert_id}'.format(api_url=self.api_url, alert_id=alert_id)
+        resource = {'description': description, 'expression': expression, 'db': miq_entity,
```
zgalor: missed this one `db` to `entity`

library/manageiq_alert.py
```diff
@@ -0,0 +1,282 @@
...
+    def create_alert(self, description, expression, miq_entity, options, enabled):
+        """Creates the alert in manageiq.
+
+        Returns:
+            Whether or not a change took place and a message describing the
+            operation executed.
+        """
+        url = '{api_url}/alert_definitions/'.format(api_url=self.api_url)
+        resource = {'description': description, 'expression': expression,
+                    'db': miq_entity, 'options': options, 'enabled': enabled}
```
zgalor: same

tests/test_manageiq_alert.py
```diff
@@ -0,0 +1,214 @@
...
+            "description": DESCRIPTION,
+            "created_on": "2017-04-04T07:42:51Z",
+            "updated_on": "2017-04-04T07:42:51Z",
+            "options": {
+                "notifications": {
+                    "delay_next_evaluation": 600,
+                    "evm_event": {}
+                }
+            },
+            "db": "ContainerNode",
```
zgalor, line 79: `db` --> `entity`

tests/test_manageiq_alert.py
```diff
@@ -0,0 +1,214 @@
...
+        "description": DESCRIPTION,
+        "created_on": "2017-04-04T07:42:51Z",
+        "updated_on": "2017-04-04T07:45:51Z",
+        "options": {
+          "notifications": {
+            "delay_next_evaluation": 60,
+            "evm_event": {}
+          }
+        },
+        "db": "ContainerNode",
```
zgalor, line 103: same here

tests/test_manageiq_alert.py
```diff
@@ -0,0 +1,214 @@
...
+
+    result = miq.create_or_update_alert(DESCRIPTION, EXPRESSION, MIQ_ENTITY, OPTIONS, ENABLED)
+    assert result == {
+        'changed': True,
+        'msg': "Successfully created alert {description}: {alert_details}".format(description=DESCRIPTION, alert_details=POST_RETURN_VALUES['created_alert']['results'])
+    }
+    miq.client.post.assert_called_once_with(
+        '{hostname}/api/alert_definitions/'.format(hostname=MANAGEIQ_HOSTNAME),
+        action='create',
+        resource={'description': DESCRIPTION, 'expression': EXPRESSION, 'db': 'ContainerNode', 'options': OPTIONS, 'enabled': ENABLED}
```
zgalor: `db` --> `entity`

zgalor (commented):
Just a few more `db` missed

library/manageiq_alert.py
dkorn: This one should be `db`. it is the format of the resource the API expects.

library/manageiq_alert.py
dkorn: same answer, should be `db`

tests/test_manageiq_alert.py
dkorn: again

tests/test_manageiq_alert.py
dkorn: again

tests/test_manageiq_alert.py
dkorn: again

library/manageiq_alert.py
zgalor: ok. cool :+1:

zgalor:
:+1:  LGTM

zgalor (approved):

library/manageiq_alert.py
```diff
@@ -0,0 +1,282 @@
...
+        """
+        url = "{api_url}/alert_definitions/{alert_id}".format(api_url=self.api_url, alert_id=alert_id)
+        try:
+            result = self.client.get(url)
+        except Exception as e:
+            self.module.fail_json(msg="Failed to get alert {description} details. Error: {error}".format(description=description, error=e))
+
+        # remove None values from expression and options dicts
+        current_expression = dict((k, v) for k, v in result['expression']['exp'].items() if v is not None)
+        current_options = dict((k, v) for k, v in result['options'].items() if v is not None)
```
cben: python 2.7+ has dict comprehension syntax
{k: v for ... if ...}

library/manageiq_alert.py
```diff
@@ -0,0 +1,282 @@
...
+
+        attributes_tuples = [(current_expression, expression), (result['db'], miq_entity), (current_options, options), (result['enabled'], enabled)]
+        for tup in attributes_tuples:
+            if tup[1] is not None and tup[0] != tup[1]:
```
cben: readabilty: prefer destructuring `for (current, desired) in tuples:` over tup[0], tup[1].

Why None check?  So things user didn't specify are untouched?  I'd like to learn once and for all if that's good ansible practice...

Like in many module we got duplicate logic in this function fetching parts from get response, vs create/update functions building json from part.
Can't we formulate this as normalize(build(parts...)) == normalize(current) ?

cben: OK, I see update/create have trivial logic - just a dict {'foo': foo,...}.
Still, consider  uilding that dict early (in main?) and passing it around.  That way functions will have less arguments, and comparisons will be 1:1 eg. `current['db']` vs `desired ['db']`

README.md
```diff
@@ -69,6 +69,13 @@ Each assigned tag must be a part of a unique category.
 An example playbook [assign_tag.yml](assign_tag.yml) is provided.  
 To unassign a tag from a resource change `state=absent`.
 
+### manageiq_alert module
+
+The `manageiq_alert` module supports adding, updating and deleting alerts in manageiq.  
+Example playbook [create_alert.yml](examples/create_alert.yml) is provided.  
+To update an existing alert pass the changed values together with the required parameters. To delete an alert change `state=absent`.
+The following ManageIQ entities supports alerts: node, vm, host, storage, cluster, ems, miq_server and middleware_server.
```
moolitayer: s/supports/support/ 
s/node/container node/  (since that's the name in ManageIQ)

examples/create_alert.yml
```diff
@@ -0,0 +1,21 @@
...
+  - name: Create an alert in ManageIQ
+    manageiq_alert:
+      miq_url: http://miq.example.com
+      miq_username: admin
+      miq_password: secret
+      miq_verify_ssl: false
+      description: Test Alert 01
+      options:
+        notifications:
+          delay_next_evaluation: 600
```
moolitayer: oh that's a bad name for that. I've created a ticket to rename this in manageiq model
https://trello.com/c/dfu1Qumi/138-rename-delay-next-evaluation-in-manageiq

library/manageiq_alert.py
```diff
@@ -0,0 +1,282 @@
...
+module: manageiq_alert
+description: The manageiq_alert module supports adding, updating and deleting alert definitions in ManageIQ.
+short_description: management of alerts in ManageIQ
+requirements: [ ManageIQ/manageiq-api-client-python ]
+author: Daniel Korn (@dkorn)
+options:
+  miq_url:
+    description:
+      - the manageiq environment url
+    default: MIQ_URL env var if set. otherwise, it is required to pass it
```
moolitayer, line 15: maybe 
s/otherwise, it is required to pass it/otherwise, it is required here/

cben:
Module name: in API we called MiwAlert "alert_definition" vs MiqAlertStatus "alert" happening to specific object at specific moment.
Should this maybe be called manageiq_alert_definition?
Are we sure we'll never add ansible module for MiqAlertStatus?

library/manageiq_alert.py
moolitayer: container node?(is that what you would pass in the api?)

library/manageiq_alert.py
```diff
@@ -0,0 +1,282 @@
...
+    description:
+      - the alert definition description in manageiq
+    required: false
+    default: null
+  entity:
+    description:
+      - the entity to base the alert on in manageiq
+    required: false
+    choices: ['node', 'vm', 'host', 'storage', 'cluster', 'ems', 'miq_server', 'middleware_server']
+    default: null
```
moolitayer: do you think we might want nil here eventually or is this how this is done in ansible?

cben: null is the YAML nil.  goes through a python None, JSON null on the wire, ruby nil and SQL null :-)
but why is this optional, can a MiqAlert work without `db`?

library/manageiq_alert.py
dkorn: Done, thanks!

README.md
dkorn: I think it should be supports
switched to 'container node', thanks!

moolitayer: deferring to English speaker, @cben?

library/manageiq_alert.py
dkorn: Changed to `(current, desired)`, great suggestion!

I'm not sure what I think about creating the dict in main, maybe on a separate function like we do for provider endpoints. Anyway, this won't be consistent with the other modules.

examples/create_alert.yml
dkorn: I agree, but as long as it is the name in manageiq not much to do

library/manageiq_alert.py
dkorn: It is the same in all modules, so I prefer consistency (unless we change all)

dkorn:
> Module name: in API we called MiwAlert "alert_definition" vs MiqAlertStatus "alert" happening to specific object at specific moment.
> Should this maybe be called manageiq_alert_definition?
> Are we sure we'll never add ansible module for MiqAlertStatus?

I don't have a clear preference here, @simon3z, @moolitayer your opinions?

library/manageiq_alert.py
dkorn: > but why is this optional, can a MiqAlert work without db?

@cben yes, In case we delete an alert we only need its name. that's why: 
```python
required_if=[
    ('state', 'present', ['description', 'expression', 'entity', 'options'])
],
```

README.md
cben: "The `manageiq_alert` module supports" LGTM.

library/manageiq_alert.py
cben: Please make sure doc says they're required except when deleting.  And sounds like `description` should always be required.

dkorn: Done

dkorn:
@cben @moolitayer, unless we rename the module (cc @simon3z) can this be merged?

library/manageiq_alert.py
```diff
@@ -0,0 +1,284 @@
...
+    entity: container_node
+    expression:
+      eval_method: dwh_generic
+      mode: internal
+    enabled: true
+    state: present
+    miq_url: 'http://localhost:3000'
+    miq_username: 'admin'
+    miq_password: '******'
+    miq_verify_ssl: False
```
moolitayer, line 97: `policy.mode = control|compliance` is missing and important to us

examples/create_alert.yml
```diff
@@ -0,0 +1,21 @@
...
+          evm_event: {}
```
cben: what does this do, is it necessary?

examples/create_alert.yml
```diff
@@ -0,0 +1,21 @@
...
+      entity: node
+      expression:
+        eval_method: dwh_generic
```
cben: a lot of MIQ internal wizardry :-(
it may be useful to have 2 examples, one for "normal"alert, one for dwh alert.  in fact the later would ideally live together with the hawkular trigger definition. and IIUC it takes 2 manageiq_provider definitions for openshift+linked dwh, I'd throw those in as well...

library/manageiq_alert.py
```diff
@@ -0,0 +1,284 @@
...
+        'miq_server': 'MiqServer', 'middleware_server': 'MiddlewareServer'
+    }
+
+    def __init__(self, module, url, user, password, miq_verify_ssl, ca_bundle_path):
+        self.module        = module
+        self.api_url       = url + '/api'
+        self.user          = user
+        self.password      = password
+        self.client        = MiqApi(self.api_url, (self.user, self.password), verify_ssl=miq_verify_ssl, ca_bundle_path=ca_bundle_path)
+        self.changed       = False
```
moolitayer, line 126: we know this isn't thread safe right?

we should asses if this is fine for ansible use case(not in this PR)
@cben @dkorn

library/manageiq_alert.py
```diff
@@ -0,0 +1,284 @@
...
+    default: 'present'
+'''
+
+EXAMPLES = '''
+# Create a new alert in ManageIQ
+  manageiq_alert:
+    description: Test Alert 01
+    options:
+      notifications:
+        delay_next_evaluation: 60
```
moolitayer: since a recent PR this will always fail. Please change to 0

library/manageiq_alert.py
moolitayer: you must pass some params here as ruby symbols (:x) or this will not work - at least not with the example provided.

moolitayer: for example `options`

examples/create_alert.yml
moolitayer: better have an example with some notifications if those are supported.
(can leave to other iteration, we will not use them for now)

library/manageiq_alert.py
moolitayer: consider changing all (not in this pr)

library/manageiq_alert.py
```diff
@@ -0,0 +1,284 @@
...
+        evm_event: {}
+    entity: container_node
+    expression:
```
moolitayer: `expression` can be either miq_expression or  hash https://github.com/moolitayer/manageiq/blob/b3af72f325ff42f2743154ce8c796b6b12084aeb/app/models/miq_alert.rb#L328-L331

The way we pass it now seems to always create a miq_expression instead of a hash.
```
#<MiqExpression:0x000000088d3f30 @exp={":mode"=>"internal", ":eval_method"=>"dwh_generic"}, @context_type=nil>
```

The reason is here:
https://github.com/manageiq/manageiq/blob/ca15956dca10d194cd683106a72c1c8e339258fa/app/controllers/api/alert_definitions_controller.rb#L9

So: we will need a way to allow both types. I'm not sure how to do in the API, talk to those guys.
We might need another field in the module evaltualy

library/manageiq_alert.py
cben: Is fine.  (1) ansible modules always run in separate process. (2) we do want ability to run multiple actions in tests, but both `main()` and tests always create new instance, run one action (create/update/etc), discard instance.

library/manageiq_alert.py
dkorn: you're confusing alerts with policies right?

examples/create_alert.yml
dkorn: It is the outcome I got when created an alert with a timeline event notification

cben (commented):
No more comments from me, deferring to @moolitayer, ping me if you'll want mora review

library/manageiq_alert.py
dkorn: Always creating miq_expression was Alberto's request. 
How do you suggest to change it on the ManageIQ side? use of dedicated flag?

library/manageiq_alert.py
dkorn: so we can't determine the time between checks anymore?

library/manageiq_alert.py
dkorn: OK

examples/create_alert.yml
dkorn: > and IIUC it takes 2 manageiq_provider definitions for openshift+linked dwh, I'd throw those in as well...

@cben Not sure what you meant here, why 2?

> better have an example with some notifications if those are supported.

@moolitayer Do you have an example of alert with notidication?

examples/create_alert.yml
cben: ah, makes sense :+1:

examples/create_alert.yml
cben: don't you need both an openshift provider and a datawarehouse provider?  I might be totally confused...
anyway not a blocker for this PR, just thoughts on examples that would be nice...

library/manageiq_alert.py
moolitayer: We can, in this specific type (`dwh_generic`) this time is enforced to always be 0

library/manageiq_alert.py
moolitayer: yes, there is something very similar here:
https://github.com/manageiq/manageiq/blob/39b1dfc35e3755ec474690bf7f3d987f566e04c7/app/controllers/api/providers_controller.rb#L92

library/manageiq_alert.py
moolitayer: right

library/manageiq_alert.py
dkorn: @moolitayer https://github.com/ManageIQ/manageiq/pull/14979

library/manageiq_alert.py
dkorn: Got it, will change the example, thanks!

simon3z:
@dkorn @moolitayer @zgalor @cben is this finalized?
@cben @moolitayer @zgalor can you review/approve?

moolitayer:
@dkorn was the expression type issue resolved? Didn't see that in the example but maybe that is out of date?

moolitayer:
@dkorn if so please let me know and I'll take it for another test.

dkorn:
@mtayer @simon3z this PR might be dependent on this change https://github.com/ManageIQ/manageiq/pull/14979, meaning the Alert module will work but the ManageIQ REST API will always create `MiqExpression` typed alerts, instead of hash typed which are needed for us (@moolitayer can elaborate).

The dependent PR might take some time to get merged since there's a migration needed to `MiqAlert`.
I guess what I'm trying to say is that it is up to us if we want this to be merged now or wait to the ManageIQ API and table changes.

simon3z:
@dkorn if the API is settled for sure and it just needs review/work of the implementation in https://github.com/ManageIQ/manageiq/pull/14979 then we can try to finalize this.

dkorn:
> @dkorn if the API is settled for sure and it just needs review/work of the implementation in ManageIQ/manageiq#14979 then we can try to finalize this.

@simon3z I think we're covered on the API side (Alberto sort of approved the "design").
I'll just add the expression_type attribute and ask for @cben and @moolitayer final review

dkorn:
@cben @moolitayer @zgalor can you please take a look at the `expression_type` addition (last commit)

library/manageiq_alert.py
```diff
@@ -160,8 +167,9 @@ def alert_update_required(self, alert_id, description, expression, miq_entity, o
         except Exception as e:
             self.module.fail_json(msg="Failed to get alert {description} details. Error: {error}".format(description=description, error=e))
 
-        # remove None values from expression and options dicts
-        current_expression = {k: v for k, v in result['expression']['exp'].items() if v is not None}
+        # remove None values from expression and options dicts, if needed
+        if expression_type == 'miq_expression':
+            current_expression = {k: v for k, v in result['expression']['exp'].items() if v is not None}
```
cben: missing else clause?

If/when it'll be possible to get expression_type back from API, should use that.

dkorn: Done.

Also added a TODO comment to use the returned `expression_type`, once supported by ManangeIQ API.

cben (approved):

dkorn:
Tested against https://github.com/ManageIQ/manageiq/pull/14979 and it worked.

@moolitayer can you give your final :+1: ?

examples/create_alert.yml
```diff
@@ -0,0 +1,22 @@
...
+        notifications:
+          delay_next_evaluation: 0
+          evm_event: {}
+      entity: node
+      expression:
+        eval_method: dwh_generic
+        mode: internal
+      expression_type: hash
+      enabled: true
+      state: present
```
moolitayer: Daniel can you make sure the alert you showed me today can be created using this example?

dkorn: `options: {}` under `expression` was missing, added it. thanks!

moolitayer (approved):
LGTM :+1: 

Please squash...

cben:
I propose we merge this, and adjust later if anything changes in manageiq PRs.

dkorn:
> I propose we merge this, and adjust later if anything changes in manageiq PRs.

I agree, merging

PR #74, opened 2017-05-07
Refactor: default state to 'present' in all modules

dkorn:
This change enforces a unified approach for the state parameter, not required and defaulted to present in all modules.
Few documentation fixes were needed.

### DIFF
--- a/library/manageiq_custom_attributes.py
+++ b/library/manageiq_custom_attributes.py
@@ -42,7 +42,7 @@ options:
         already exist, or update the custom attributes if the associated data
         is different
       - On absent, it will delete the custom attributes from the entity if exist
-    required: false
+    required: False
     choices: ['present', 'absent']
     default: 'present'
   custom_attributes:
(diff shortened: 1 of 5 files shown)


dkorn:
@cben @enoodle @zgalor please review

zgalor:
:+1: LGTM
I suggest you also update the example playbooks, and maybe allow changing state via CLI without having to edit the playbook.

in the playbook:
`state: "{{state | default('present')}}"`

from CLI add `-e state=absent`

cben:
LGTM.
Nit: in the docs, you mostly have lowercase `required: true`, so for consistency use lowercase `required: False` here too.  (yaml itself doesn't care: http://yaml.org/type/bool.html)

> examples ... maybe allow changing state via CLI:
> ```
> state: "{{state | default('present')}}"
> ```

:+1:.  Could you get away with :

```
state: "{{state}}"
```

relying on the module's default?  IIRC not, that would provide an empty string, you probably need:

```
state: "{{state | default(omit)}}"
```

zgalor (approved):

dkorn:
> Nit: in the docs, you mostly have lowercase required: true, so for consistency use lowercase required: False here too. (yaml itself doesn't care: http://yaml.org/type/bool.html)

@cben Unfortunately, there is no real consistency in the docs to maintain :/ 
I think I'll dedicate a separate PR for this.

cben (approved):

PR #75, opened 2017-05-26
[WIP] ManageIQ Utils: moving all shared code to manageiq_utils

dkorn:
The utils module is a best practice/guideline of modules part of the `ansible/ansible` repo.

In addition:
* making the required changes in manageiq_user module
* adding the GNU General Public License V3 to the files headers
(separate PR#60 takes care of the COPYING file)

TODO (dkorn): Fix tests

### DIFF
--- a/library/manageiq_user.py
+++ b/library/manageiq_user.py
@@ -1,75 +1,77 @@
 #!/usr/bin/python
+#
+# (c) 2017, Daniel Korn <korndaniel1@gmail.com>
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+ANSIBLE_METADATA = {'metadata_version': '1.0',
+                    'status': ['preview'],
+                    'supported_by': 'community'}
 
 
 DOCUMENTATION = '''
----
+
 module: manageiq_user
-description: The manageiq_user module supports adding, updating and deleting users in ManageIQ.
-short_description: management of users in ManageIQ
-requirements: [ ManageIQ/manageiq-api-client-python ]
+
+short_description: Management of users in ManageIQ
+extends_documentation_fragment: manageiq
+version_added: '2.4'
 author: Daniel Korn (@dkorn)
+description:
+  - The manageiq_user module supports adding, updating and deleting users in ManageIQ.
+
 options:
-  miq_url:
-    description:
-      - the manageiq environment url
-    default: MIQ_URL env var if set. otherwise, it is required to pass it
-  miq_username:
-    description:
-      - manageiq username
-    default: MIQ_USERNAME env var if set. otherwise, it is required to pass it
-  miq_password:
-    description:
-      - manageiq password
-    default: MIQ_PASSWORD env var if set. otherwise, it is required to pass it
   name:
     description:
-      - the unique userid in manageiq, often mentioned as username
+      - The unique userid in manageiq, often mentioned as username
     required: true
     default: null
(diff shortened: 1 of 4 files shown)


library/manageiq_user.py
```diff
@@ -83,180 +101,140 @@
...
 
-        user_id = self.find_user_by_userid(userid)
-        if user_id:  # user already exist
-            return self.update_user_if_required(user_id, userid, username, group_id, password, email)
+        user = manageiq.find_collection_resource_by('users', userid=userid)
+        if user:  # user already exist
+            return self.update_user_if_required(manageiq, user, userid, username, group['id'], password, email)
         else:
-            return self.create_user(userid, username, group_id, password, email)
+            return self.create_user(manageiq, userid, username, group['id'], password, email)
```
cben: I'd maybe give `manageiq` to the constructor, and use `self.manageiq` everywhere.
That way you don't have to pass it through from method to method.

library/manageiq_user.py
```diff
@@ -83,180 +101,140 @@
...
+        user = manageiq.find_collection_resource_by('users', userid=userid)
+        if not user:  # user doesn't exist
             return dict(
                 changed=self.changed,
                 msg="User {userid} does not exist in manageiq".format(userid=userid))
         try:
-            url = '{api_url}/users/{user_id}'.format(api_url=self.api_url, user_id=user_id)
-            result = self.client.post(url, action='delete')
+            url = '{api_url}/users/{user_id}'.format(api_url=manageiq.api_url, user_id=user['id'])
+            result = manageiq.client.post(url, action='delete')
```
cben: I think it's worth having `manageiq.post` wrapper that takes a relative url like `users/{id}` and prepends `api_url` automatically.

library/manageiq_user.py
```diff
@@ -83,180 +101,140 @@
...
         Returns:
             the created user id, name, created_on timestamp,
             updated_on timestamp, userid and current_group_id
         """
-        if not self.user_update_required(user_id, userid, username, group_id, email):
+        if not self.user_update_required(user, username, group_id, email):
             return dict(
                 changed=self.changed,
                 msg="User {userid} already exist, no need for updates".format(userid=userid))
+        url = '{api_url}/users/{user_id}'.format(api_url=self.api_url, user_id=user.id)
```
cben: s/self.api_url/manageiq.api_url/ ?

cben (commented):
Very nice!  The signal-to-noise in the module is much better now.  Few small suggestions.

Cross-ref upstream PR: https://github.com/ansible/ansible/pull/25106

library/manageiq_user.py
dkorn: I agree, but I prefer to add it in a separate PR, this one is quite a big change already.

library/manageiq_user.py
dkorn: Done, thanks!

library/manageiq_user.py
dkorn: Interesting suggestion, I'll give it a try

library/manageiq_user.py
cben: didn't push?

dkorn: @cben soon :)

library/manageiq_utils.py
```diff
@@ -0,0 +1,80 @@
...
+    """
+
+    def __init__(self, module):
+        """
+            module         - an AnsibleModule instance
+            url            - manageiq environment url
+            user           - the username in manageiq
+            password       - the user password in manageiq
+            miq_verify_ssl - whether SSL certificates should be verified for HTTPS requests
+            ca_bundle_path - the path to a CA_BUNDLE file or directory with certificates
```
cben: is this is a remainder from earlier version where this constructor took all these params separately?

setup.py
```diff
@@ -9,6 +9,6 @@
     package_dir={'': 'library'},
     py_modules=["manageiq_provider", "manageiq_policy_assignment",
                 "manageiq_custom_attributes", "manageiq_user",
-                "manageiq_tag_assignment"],
+                "manageiq_tag_assignment", "manageiq_utils"],
```
cben, line 12: Tests fail on `import manageiq as manageiq_utils`.  Not sure why, but perhaps dropping the `_utils` here will help?

cben:
A similar refactoring has been done as the modules were rewritten into Ansible itself:
https://github.com/ansible/ansible/blob/devel/lib/ansible/module_utils/manageiq.py

We significantly rewrote all modules anyway, with incompatible interface changes, so we didn't even try to keep this repo in sync.  Closing.

PR #76, opened 2017-06-08
Create Openshift provider with Prometheus endpoint

zgalor:
Replaced `metrics: True` with default hawkular to generic endpoint terminology:
`monitoring: 'hawkular' / 'prometheus' / None`

endpoint params replaced accordingly:
`hawkular_hostname` => `monitoring_hostname`
`hawkular_port` => `monitoring_port`

Also added a creation example for prometheus in examples/add_openshift_provider.yml
and a test in tests/test_manageiq_provider.py

Temporary hack - No validation for prometheus endpoint authentication until the is supported by manageiq.

**Note**: this changes the module API and older playbooks with hawkular endpoint will not be supported from now on


zgalor:
@cben @dkorn @moolitayer please review

examples/add_openshift_provider.yml
```diff
@@ -19,5 +19,25 @@
...
+      name: 'Openshift02'
+      provider_type: 'openshift-origin'
+      state: 'present'
+      provider_api_hostname: 'os02.example.com'
+      provider_api_port: '8443'
+      provider_api_auth_token: '******'
+      metrics: true
+      metrics_type: 'prometheus'
+      prometheus_hostname: 'prometheus01.example.com'
+      prometheus_port: '80'
```
cben: So this needs 4 parameters, and could be 2 or 3.
`metrics: true` and `metrics_type: 'prometheus'` are IMO redundant.  I propose combining to `metrics: 'prometheus'`.

In theory, both could be infered from the fact you provided `prometheus_hostname:` and `prometheus_port:`, but IIRC we added separate `metrics: true` boolean to be able to ~~specify it but omit host/port which had a sensible default(?)~~

@dkorn @enoodle what do you think?

zgalor: So how about:
metrics : 'hawkular' / 'prometheus' / none (by default)
Then of course we could have `metrics_port, metrics_hostname` regardless of the venue.

But this is not backwards compatible with old playbooks. do we want it to be?

cben: Actually I don't see any trace of hawkular host/port being optional.

cben (commented):
Code LGTM. (aside from the interface metrics/metrics_type discussion)

examples/add_openshift_provider.yml
moolitayer: I like best the option of having `hawkular_hostname`, `hawkular_port` OR `prometheus_hostname`, `prometheus_port`. Bad combinations should fail.
I would not touch `metrics: true` in this PR. 
I don't think we should be concerned with backward compatibility at this stage.

library/manageiq_provider.py
```diff
@@ -531,16 +551,18 @@ def main():
...
+            monitoring_port=dict(required=False),
             initiate_refresh=dict(required=False, type='bool', default=True),
             validate_provider_auth=dict(required=False, type='bool', default=True)
         ),
         required_if=[
             ('provider_type', 'openshift-origin', ['provider_api_hostname', 'provider_api_port', 'provider_api_auth_token']),
             ('provider_type', 'openshift-enterprise', ['provider_api_hostname', 'provider_api_port', 'provider_api_auth_token']),
-            ('metrics', True, ['hawkular_hostname', 'hawkular_port']),
+            ('monitoring', 'hawkular', ['monitoring_hostname', 'monitoring_port']),
+            ('monitoring', 'prometheus', ['monitoring_hostname', 'monitoring_port']),
```
zgalor, line 565: I am not so happy with writing this validation twice - once for prometheus and once for hawkular, but I couldn't find a way in `required_if` to state "not None". Suggestions are welcome...

zgalor:
After a consult with @moolitayer I decided to rename the endpoint parameters to `monitoring`:
```
monitoring: prometheus / hawkular / None
monitoring_hostname
monitoring_port
```

This aligns with the new naming convention in manageiq, and more generic for future usage of logging etc.

@cben @yaacov @dkorn @moolitayer  please review

examples/add_openshift_provider.yml
```diff
@@ -2,17 +2,36 @@
...
-      hawkular_hostname: 'hawkular01.example.com'
-      hawkular_port: '443'
+      monitoring: 'hawkular'
+      monitoring_hostname: 'hawkular01.example.com'
+      monitoring_port: '443'
+      miq_url: 'http://miq.example.com'
+      miq_username: 'admin'
+      miq_password: '******'
+      miq_verify_ssl: false
+      provider_verify_ssl: false
```
yaacov, line 20: I think it's nicer if the `provider_verify_ssl` will be together with the `provider_api_hostname` and `api_port`

library/manageiq_provider.py
cben: I don't think there is.
https://github.com/ansible/ansible/blob/b69c7f50d02dac3e24bf1927ebdc933475da93b7/lib/ansible/module_utils/basic.py#L1635 checks `== val`.
Are you unhappy with textual duplication (LGTM) or the fact we'll have to remember to extend it if we add another?

required_if is not magical.
If you want you can check the condition yourself and exit with error (like done below for  `['miq_url', 'miq_username', 'miq_password']`)...

cben:
If "monitoring" term can include logging, will we at some point need 2 monitoring endpoints of different kinds?

zgalor:
@cben Probably not. More likely that each will reside in a different manager, @moolitayer keep me honest here.

examples/add_openshift_provider.yml
zgalor: I agree, but this is only an example, and also irrelevant to this PR

library/manageiq_provider.py
zgalor: I am unhappy with both, but then again it might be clearer to leave it like this. I the list or choices grows longer, we should refactor.

library/manageiq_provider.py
```diff
@@ -531,16 +551,18 @@ def main():
             provider_region=dict(required=False, type='str'),
             access_key_id=dict(required=False, type='str', no_log=True),
             secret_access_key=dict(required=False, type='str', no_log=True),
-            metrics=dict(required=False, type='bool', default=False),
-            hawkular_hostname=dict(required=False),
-            hawkular_port=dict(required=False),
+            monitoring=dict(required=False, default=None,
+                       choices=['hawkular', 'prometheus', None]),
+            monitoring_hostname=dict(required=False),
```
cben, line 556: Could add `type='str'`, also missing on many others. Not necessarily in this PR.

library/manageiq_provider.py
```diff
@@ -577,8 +598,10 @@ def main():
     if state == 'present':
         if provider_type in ("openshift-enterprise", "openshift-origin"):
             endpoints = [manageiq.generate_auth_key_config('default', 'bearer', hostname, port, token, provider_verify_ssl, provider_ca_path)]
-            if module.params['metrics']:
-                endpoints.append(manageiq.generate_auth_key_config('hawkular', 'hawkular', h_hostname, h_port, token, provider_verify_ssl, provider_ca_path))
+            if monitoring:
+                monitoring_hostname =  module.params['monitoring_hostname']
+                monitoring_port = module.params['monitoring_port']
+                endpoints.append(manageiq.generate_auth_key_config(monitoring, monitoring, monitoring_hostname, monitoring_port, token, provider_verify_ssl, provider_ca_path))
```
cben: I'd pass these by name
```
generate_auth_key_config(
    role=monitoring,
    authtype=monitoring,
    ...
```

tests/test_manageiq_provider.py
```diff
@@ -86,7 +88,7 @@
 }
 
 GET_RETURN_VALUES = {
-    'openshift_without_hawkular': {
+    'openshift_without_metrics': {
```
cben: consider s/openshift_without_metrics/openshift_without_monitoring/g.

more importantly, there is `miq_ansible_module.params = {'metrics': True}` on L274

cben (commented):
Looks good.

library/manageiq_provider.py
zgalor: since we have "choices", this seems obsolete

library/manageiq_provider.py
zgalor: Very well.
So I guess this should be done for other calls, right?

tests/test_manageiq_provider.py
zgalor: nice catch. thanks

library/manageiq_provider.py
cben: Your call ;-).
Personally I find I can't read the current `generate_auth_key_config()` call without going to see its definition.  I suppose if you change this, should also change the 2 other endpoint-generating calls.

tests/test_manageiq_provider.py
zgalor: I wonder how the test passed with the `metrics: true` param. I will fix the test

zgalor:
I think this is ready now.
@cben @moolitayer please take a final look

cben (approved):

cben:
@dkorn I expect you have new priorities :) I'm starting to merge without reviews from you; let us know if you want to stay in the loop.

cben:
@zgalor please edit PR description according to current interface.

zgalor:
done.

moolitayer:
:tada:

PR #77, opened 2017-06-19
Updated cm-ops deployment playbook

zgalor:
Playbook for cm-ops deployment test:
Create manageiq users
Create openshift provider with prometheus endpoint
Add custom attributes to provider


zgalor:
Irrelevant

PR #79, opened 2017-06-29
Add manageiq_fetch_key and manage_join_region modules

itewk:
Created two new modules for fetching key and joining a region as part of the CFME hack a thon. Let me know if you need more in pull request to accept. 

As a side note, additionally it would be great to get this module hosted on the rhtconsulting github to get it even more visibility.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ To view a module documentation execute:
 
 ### manageiq_provider module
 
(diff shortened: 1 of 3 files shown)


cben:
Yay, more modules :)

We started the process of contributing modules from this repo to ansible itself (https://github.com/ansible/ansible/issues/25106), which should get maximum visibility.

Consider submitting directly to ansible instead of this repo too, like @abraverm does in https://github.com/ansible/ansible/pull/25865.
Actually, this sounds very overlapping (?) with that module!  Please compare the efforts...

PR #84, opened 2017-07-22
add_tag.yml

pavelzag:
This PR adds the add_tag.yml example

### DIFF
--- a/examples/add_tag.yml
+++ b/examples/add_tag.yml
@@ -1,0 +1,17 @@
+---
+- hosts: localhost
+  tasks:
+  - manageiq_tag_assignment:
+      miq_password: '******'
+      miq_url: 'http://miq.example.com'
+      miq_username: 'admin'
+      miq_verify_ssl: false
+      resource: provider
+      resource_name: openshift_provider
+      state: present
+      tags:
+      - {category: environment, name: qa}
+      - {category: department, name: accounting}
+    name: Create a tag in ManageIQ
+    register: result
+  - {debug: var=result}
--- a/examples/add_tag.yml
+++ b/examples/add_tag.yml
@@ -10,8 +10,8 @@
       resource_name: openshift_provider
       state: present
       tags:
-      - {category: environment, name: qa}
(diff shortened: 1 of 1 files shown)


examples/add_tag.yml
```diff
@@ -0,0 +1,17 @@
...
+      miq_verify_ssl: false
+      resource: provider
+      resource_name: openshift_provider
+      state: present
+      tags:
+      - {category: environment, name: qa}
+      - {category: department, name: accounting}
+    name: Create a tag in ManageIQ
+    register: result
+  - {debug: var=result}
```
cben: can you stick to yaml-style braceless indentation / `-` syntax for consistency?

cben (commented):
Thanks!

examples/add_tag.yml
pavelzag: Sure, will fix

examples/add_tag.yml
```diff
@@ -10,8 +10,8 @@
       resource_name: openshift_provider
       state: present
       tags:
-      - {category: environment, name: qa}
-      - {category: department, name: accounting}
+      - category: environment, name: qa
+      - category: department, name: accounting
```
cben: Thanks but I'm afraid this comma is not legal YAML (ansible complains).  I believe it should be:
```
       tags:
       - category: environment
         name: qa
       - category: department
         name: accounting
```

cben:
Merging with 2 more changes:

- YAML parser doesn't like the `*****`, added quotes [2e9ea36feb609105d35c7cd844bf23bd113aac73]
  (one should put real password there before running the example but let's avoid the confusing error message)

- Renamed example add_tag.yml -> assign_tag.yml [5391d98498eeb0e3060c49fadd680387ce8bc0dc]

  This example assigns an existing tag;
  at some point in future we'll probably also have a module to define new tags, so reserving `add_tag` for that.
  
  Plus README.md already has a link to assign_tag.yml (yes, before the example existed!)

PR #86, opened 2017-12-08
Initial commit for ovirt provider

snecklifter:
Hello,

I recently had to create a number of providers using the uri module which was a bit painful.

Here is my first stab at adding RHEV/ovirt capability. I am not much of a coder so much of it is copy/paste. The relevant code which successfully created a RHEV provider from the uri module is:
```

- name: Create RHEV Provider
  uri:
    url: "https://{{ hostvars[groups['cfme_rhev'][0]]['inventory_hostname'] }}/api/providers"
    method: POST
    user: "{{ vault_cfme_user }}"
    password: "{{ vault_cfme_password }}"
    body: 
      type: "ManageIQ::Providers::Redhat::InfraManager"
      name: "{{ hostvars[groups['cfme_rhev'][0]]['rhev_name'] }}"
      hostname: "{{ hostvars[groups['rhev'][0]]['inventory_hostname'] }}"
      credentials:
        userid: "{{ vault_rhev_user }}"
        password: "{{ vault_rhev_password }}"
    status_code: 200
    body_format: json
    validate_certs: no
```
I don't have access to test this any more unfortunately but perhaps its a start?

Constructive criticism appreciated. I'll progress to OpenStack, Tower and Satellite if response is generally positive.

### DIFF
--- a/examples/add_ovirt_provider.yml
+++ b/examples/add_ovirt_provider.yml
@@ -1,0 +1,19 @@
+---
+- hosts: localhost
+
+  tasks:
+  - name: Add RHEV or ovirt provider to ManageIQ
+    manageiq_provider:
+      name: 'RHEV01'
+      provider_type: 'ovirt'
+      state: 'present'
+      provider_api_hostname: 'https://rhev.example.com'
+      provider_api_user: 'admin@internal'
+      provider_api_password: '******'
+      miq_url: 'http://miq.example.com'
+      miq_username: 'admin'
+      miq_password: '******'
+      miq_verify_ssl: false
+    register: result
+
+  - debug: var=result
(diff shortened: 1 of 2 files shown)


cben:
Hi, thanks!
Sorry we haven't kept the README clear - this repo isn't really maintained,
instead we've got most stuff here (with some interface changes) into
ansible/ansible itself.

Code LGTM but the goal should be to get similar addition into ansible.  Do
you want to send a PR there, or do you want us to carry this on?
cc @yaacov

I don't have an ovirt to test against but I work with people who do :-)

>

yaacov:
Hi, thanks !

> the goal should be to get similar addition into ansible.

you can take a look at current code here:
https://github.com/ansible/ansible/blob/devel/lib/ansible/modules/remote_management/manageiq/manageiq_provider.py

snecklifter:
Thanks, I'll send an updated PR shortly.

cben:
ref https://github.com/ansible/ansible/pull/33751

